Issue #21, opened 2020-08-11
Warning message There are 1 result in use. The connection will be released when they are closed


Robert-Krajcik:
This occurs in module DB.R

See the following discussion on StackOverflow
https://stackoverflow.com/questions/51213886/disconnect-dbi-rsqlite-within-a-function-in-r

SQL queries are typically a three-step process (ignoring connection management):

- send the query, accept the returned "result" object
- fetch the result using the "result" object
- clear the "result" object

The third step is important, as uncleared it represents resources that are being held for that query. Some database connections do not permit multiple simultaneous uncleared results, just one query at a time.

Instead of coding
```{r}
con <- dbConnect(RSQLite::SQLite(), "./risk_assessment_app.db")
q<-dbSendQuery(con, "select * from MaintenanceMetrics")
q<-dbFetch(q)
q
dbDisconnect(con)
```

It should be coded using dbClearResult() **before** dbDisconnect()
```{r}
con <- dbConnect(RSQLite::SQLite(), "./risk_assessment_app.db")
res<-dbSendQuery(con, "select * from MaintenanceMetrics")
q<-dbFetch(res)
q
dbClearResult(res)
dbDisconnect(con)
````
**or**
use  dbGetQuery() like this:
```{r}
con <- dbConnect(RSQLite::SQLite(), "./risk_assessment_app.db")
q<-dbGetQuery(con, "select * from MaintenanceMetrics")
dbDisconnect(con)
```

Robert-Krajcik:
The second commit addresses
`Warning in result_fetch(res@ptr, n = n) :
  SQL statements must be issued with dbExecute() or dbSendStatement() instead of dbGetQuery() or dbSendQuery()`

Robert-Krajcik:
Resolved with fix #24 Db fun update

PR #22, opened 2020-08-11
Fix for issue #21 Warning in connection_release(conn@ptr)

Robert-Krajcik:


Robert-Krajcik:
I thought this fixed the problem, but it was only because the database was loaded.
I'm going to close this and try again later.

PR #29, opened 2020-08-19
update setup.R and utils.R to use MySQL

Robert-Krajcik:
The code works, but if you don't have mysql server on your machine first, it won't


Robert-Krajcik:
Writing a vignette instead.

Issue #31, opened 2020-08-30
Remove code duplication


marlycormar:
The file `Server/communityusage_metrics.R` has hard-coded each month of the last year to display the chart "NUMBER OF DOWNLOADS IN PREVIOUS 11 MONTHS". It would be a good idea to remove this code duplication.

https://github.com/pharmaR/risk_assessment/blob/f1b7b8a6e0b3c8ceb337d72cb5036f164bdc0343/Server/communityusage_metrics.R#L111-L185

Robert-Krajcik:
I have some test code that seems to work:
```{r}
ver_release <- c(letters[1:11])
position <- 1:11
values <- list(ver_release = ver_release, position = position)
plotlines <- map2(values$ver_release, values$position, function(.x, .y) 
            list(label = list(text = .x), color = "#FF0000", width = 2, value = .y) )
```

Robert-Krajcik:
Modified the code in communityusage_metrics.R
```{r}
        hc_xAxis(
          title = list(text = "Months"),
          
          plotLines = map2(values$riskmetrics_cum$ver_release, values$riskmetrics_cum$position,
                function(.x, .y)  list(label = list(text = .x), color = "#FF0000", width = 2, value = .y) )
          
        ) %>%
```
Comparing it with the output from the redundant cum_report.R we get the same output

PR #39, opened 2020-09-03
remove code duplication in communityusage_metrics

Robert-Krajcik:
Working towards Remove code duplication #31
Next step might be to replace cum_report.R


marlycormar:
@Robert-Krajcik, this makes me so happy!

Issue #8, opened 2020-08-05
Implement package version functionality


marlycormar:
The app only computes metrics on the latest version of the given package, ignoring the version selected in the dropdown.

Robert-Krajcik:
I can confirm that if you use 0.0.0 for the version like this

`
package | version
-- | --
ggplot2 | 0.0.0
gt | 0.0.0
dplyr | 0.0.0
dplyr | 0.0.0
dbplyr | 0.0.0`

The latest version is always selected.

![image](https://user-images.githubusercontent.com/62262182/90658776-f589ed80-e211-11ea-980f-6f6671e8f6e2.png)

Robert-Krajcik:
with changes implemented so far, we display a note that indicates
the information extracted from the package will always be the latest version.
So let's put that in!
![image](https://user-images.githubusercontent.com/62262182/91727916-5dc0c380-eb70-11ea-87c2-b799416115b4.png)

in dbupload.R this piece of code grabs the latest version from `https://cran.r-project/web/packages`
```{r, eval=FALSE}
      webpage <-
        read_html(paste0(
          "https://cran.r-project.org/web/packages/",
          package_name
        ))
```
...
```{r, eval=FALSE}
      ver_html <- html_nodes(webpage, 'td')
      ver<-html_text(ver_html)
      for(i in 1:length(ver)){
        if(!is.na(ver[i])){
          if(ver[i] == "Version:"){
            ver<-ver[i+1]
          }
        }
      }
```

Robert-Krajcik:
Further down in dbupload.R there's code that provides (what appears to be) all available versions
I am showing both the code to extract the latest version and all versions here.
```{r)
package_name <- "ggplot2"
webpage <-
  read_html(paste0(
    "https://cran.r-project.org/web/packages/",
    package_name
  ))

str(webpage)

ver_html <- html_nodes(webpage, 'td')
ver<-html_text(ver_html)
str(ver)
```
` chr [1:52] "Version:" "3.3.2" "Depends:" "R (= 3.2)" "Imports:" ...`

```{r}
pkg_html <- read_html(paste0("https://github.com/cran/", package_name, "/tags"))
pkg_nodes_v <- html_nodes(pkg_html, 'h4')
pkg_text_v <- html_text(pkg_nodes_v)
pkg_text_v <- str_split(pkg_text_v,"\n")
pkg_vers <- c()
for (i in 1:length(pkg_text_v)) { 
[...]

PR #41, opened 2020-09-04
replace version with latest version in Packageinfo

Robert-Krajcik:
This is an addon to #8 Implement Package version functionality
It will display the latest version of the package instead of the one provided in the packages.csv file


Robert-Krajcik:
superseded by #52 Fix Versioning Number

Issue #42, opened 2020-09-04
Selecting Testing Metrics tab returns error message


Robert-Krajcik:
This is a bug, regardless of how we recode the testing_metrics.R and/or the tm_report.R

```{r, eval=FALSE}
Warning in .testLength(param = num, len = 1, arg = arg) :
  NAs introduced by coercion
Warning: Error in if: missing value where TRUE/FALSE needed
  99: .testInterval
  98: amAngularGauge
  95: func
  82: origRenderFunc
  81: output$test_coverage
   1: runApp
```

marlycormar:
Related to https://github.com/pharmaR/risk_assessment/issues/45.

Robert-Krajcik:
There's another way to do this.
PR #63 sets the value to zero so the amAngularGauge appears.

PR #43, opened 2020-09-04
fix for test metrics error message

Robert-Krajcik:
fix for issue #42 Selecting Testing Metrics Tab returns error message
fixed code in both testing_metrics.R and tm_report.R


Robert-Krajcik:
Maybe this really isn't necessary

Issue #49, opened 2020-09-10
Loading a non-existent package yields many un-helpful warning messages.


ghost:
I had a non-existent package () in my packages.csv file
From the console:

```{r, eval=FALSE)
> runApp()
[1] "Log file set to loggit.json"

Listening on http://127.0.0.1:4108
{"timestamp": "2020-09-10T16:39:53-0400", "log_lvl": "ERROR", "log_msg": "Error in extracting general info of the package Cumulonimbus info Error in open.connection(x__COMMA__ __DBLQUOTE__rb__DBLQUOTE__): HTTP error 404.__LF__", "app": "fileupload-webscraping"}
Warning in error_handler(x, ...) :
  no available scoring algorithm for metric of class "pkg_metric_error", returning default score of 0.
Warning in error_handler(x, ...) :
  no available scoring algorithm for metric of class "pkg_metric_error", returning default score of 0.
Warning in error_handler(x, ...) :
  no available scoring algorithm for metric of class "pkg_metric_error", returning default score of 0.
Warning in error_handler(x, ...) :
  no available scoring algorithm for metric of class "pkg_metric_error", returning default score of 0.
Warning in error_handler(x, ...) :
  no available scoring algorithm for metric of class "pkg_metric_error", returning default score of 0.
Warning in error_handler(x, ...) :
  no available scoring algorithm for metric of class "pkg_metric_error", returning default score of 0.
Warning in error_handler(x, ...) :
  no available scoring algorithm for metric of class "pkg_metric_error", returning default score of 0.
Warning in error_handler(x, ...) :
  no available scoring algorithm for metric of class "pkg_metric_error", returning default score of 0.
Warning in error_handler(x, ...) :
  no available scoring algorithm for metric of class "pkg_metric_error", returning default score of 0.
Warning in error_handler(x, ...) :
  no available scoring algorithm for metric of class "pkg_metric_error", returning default score of 0.
Warning in for (i in 1:length(args)) { :
  closing unused connection 3 (https://cran.r-project.org/web/packages/Cumulonimbus)
{"timestamp": "2020-09-10T16:39:55-0400", "log_lvl": "ERROR", "log_msg": "Error in extracting cum metric info of the package: Cumulonimbus info Error in open.connection(x__COMMA__ __DBLQUOTE__rb__DBLQUOTE__): HTTP error 404.__LF__", "app": "fileupload-webscraping"}
```
I'm thinking it would be better to compare against the packages the user has already installed.
[...]

PR #50, opened 2020-09-11
Package exists

Robert-Krajcik:
Resolves issue #49 -- a much cleaner way of checking if the package exists.
The current method is to just throw it out there and see if it generates a 404 message.

Also increased the default height of the main tab from 635px to 900px


Robert-Krajcik:
I'm going to close this one for now and maybe carry it over to #52

Issue #54, opened 2020-09-14
Integration with riskmetric


marlycormar:
Integration with riskmetric
- [x] Remove hardcoded commit number.
- [x] Remove webscrapping code and obtain any needed information from `riskmetric`.
- [x] Add new metrics on the 'Maintance Metrics' tab accordingly.
- [x] Fix broken code given this update.
- [x] Related issues: [#49](https://github.com/pharmaR/risk_assessment/issues/49), [#45](https://github.com/pharmaR/risk_assessment/issues/45), [#42](https://github.com/pharmaR/risk_assessment/issues/42),

Robert-Krajcik:
Just an FYI,
Looking through the community usage metrics
It is possible to collect more than 12 months of download data

365 days worth of download information are returned to
`downloads_1yr_br_i <- pkg_ref(package_name)$downloads`
because the pgk_ref function only collects 365 days

The riskmetric package 

```{r, eval = FALSE}
pkg_ref_cache.downloads <- function(x, ...) {
  cran_downloads(x$name, from=Sys.Date()-365, to=Sys.Date())
}
```
If we replaced it with the `cran_downloads()` function.
for example
```{r, eval=FALSE)
month <- "October 2016"
from <- as.Date(paste0('01 ', month), format="%d %B %Y")
downlds <- cranlogs::cran_downloads(package_name, from=from, to=Sys.Date())
```
We could get several years worth of data.
Now we can show from the earliest package date to the date the version we are interested in became available.

AARON-CLARK:
Nice find Robert. So then you could display radiobuttons for the user to select 1 year, 2 years, 5 years, or all of time. Something like that?

Robert-Krajcik:
I hadn't thought about radio buttons, but them or may a slider would be useful.
First I'd grab the pub date of the package version the user is interested in
Then figure out how far back to go.
Then slot into months and graph it.
It looks like we can use the `riskmetric pkg_ver()` function instead of web-scraping or reading the DESCRIPTION file
But I need to test that some more.

Robert-Krajcik:
Sorry meant to write `pkg_ref()`

aclark02-arcus:
Super stale issue. Closing!

PR #56, opened 2020-09-15
remove hardcoded commit number in setup.R

Robert-Krajcik:


Robert-Krajcik:
part of resolving issue #54 integration with riskmetric

marlycormar:
Hi @Robert-Krajcik, did you see any issues after getting rid of this commit?

marlycormar:
Tested the app and removing the hard-coded commit doesn't seem to be causing any issues.

Issue #52, opened 2020-09-14
Allow other pkg versions besides most recent version


marlycormar:
Fix versioning number
- [ ] The version dropdown should display the version(s) uploaded by the user (not the most recent one of the package).
- [ ] The db should be able to store different versions of the same package. Hence, some tables (like the Packageinfo one) should be modified accordingly (e.g. Packageinfo should have a primary key - package name - and a secondary key - package version). The `update` and `insert` may need to be updated.

Robert-Krajcik:
How can we tell if the version provided in the *.csv file is not mistyped?
Can we assume the version uploaded by the user is the one they have installed?
Or do we need to pull the version history out of CRAN?

Robert-Krajcik:
Overall plan:

- [ ] Given `uploadpackage.R` reads *.csv file into df pkgs_file
- [ ] Verify that the package name exists on CRAN
- [ ] Verify that the version exists and is available using the `versions` package function `available.versions()`
- [ ] If either test fails, issue message and delete the row in pkgs_file and reset the row numbers
- [ ] Otherwise, continue but with the following changes.
- [ ] Multiple versions per package will be allowed, not just the latest one. 
- [ ] Version dropdown in sidebar will display the version(s) uploaded by the user.

Robert-Krajcik:
In dbupload.R, information on the latest version is web-scraped.
```{r, eval=FALSE}
        read_html(paste0(
          "https://cran.r-project.org/web/packages/",
          package_name
        ))
```
However, if the package_version read is not equal to the current version, we can do something else
`if (ver != package_version) signalCondition(e)`
and use the R6Class function `description()` to get information about the package that is currently installed

```{r, eval=FALSE}
          if (file.exists(paste(i, "/", package_name, sep = "")) == TRUE) {
            i <- paste0(i, "/", package_name)
            d <- description$new(i)
            title <- d$get("Title")
            ver <- d$get("Version")
            desc <- d$get("Description")
            desc <- gsub("'", "", desc) # remove single quotes
            main <- d$get("Maintainer")
            auth <- d$get("Author")
            lis <- d$get("License")
            pub <- d$get("Packaged")
```
So it is possible to have more than one version of a package.
![image](https://user-images.githubusercontent.com/62262182/93358402-6a4f4800-f80f-11ea-8502-f60bc46aa37f.png)

[...]

PR #57, opened 2020-09-18
Fix versioning

Robert-Krajcik:
ref: #52 -- This is a work in progress

### DIFF
--- a/Modules/dbupload.R
+++ b/Modules/dbupload.R
@@ -108,7 +108,6 @@ get_packages_info_from_web <- function(package_name, package_version) {
     error = function(e) {
       print(paste("in error function for",package_name,"version",package_version))
       if (package_name %in% rownames(installed.packages()) == TRUE) {
-        print("package is in installed.packages")
         for (i in .libPaths()) {
           if (file.exists(paste(i, "/", package_name, sep = "")) == TRUE) {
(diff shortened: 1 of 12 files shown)


marlycormar:
Hi @Robert-Krajcik, you have made good progress here. Let me know if you'd like me to review some part or take over a portion of it.

Robert-Krajcik:
Hi @marlycormar 
I can see that the community metrics goes across versions of a package,
so we could have it report/graph **up to** the version selected.
(the note about 11 months of history is hardcoded and will need updating)

But the thing I am grappling with right now is,
how do we influence the version in pkg_assess()?
Assume:
`package_name <- "glue"`
and it is version 1.3.2
Just modifying the version here seems to make no difference

```{r, eval=FALSE}
package_riskmetric1a <<-
  pkg_ref(package_name) %>%
  mutate(version = "1.4.2") %>%
  as_tibble() %>%
  pkg_assess() %>%
  pkg_score() %>%
  mutate(risk = summarize_scores(.))
```

Robert-Krajcik:
Or maybe we just have the database tables and logic set up for multiple versions, and worry about riskmetric assessment later?

Robert-Krajcik:
Replaced by #63

PR #58, opened 2020-09-20
Explain code structure

marlycormar:
File was mistakenly deleted. Putting it back. 

It is still a work in progress...

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -98,10 +98,10 @@ Non-R assets for the application: images and a JavaScript helper file to app rea
 
 ### Files created by the app
 
-- `loggit.json`: Log file created each time the application is ran. It contains information about the uploaded packages and application errors. For example, when the user uploads the example csv file `Upload_file_structure.csv`, the following line is added to the `loggit.json` file:
+- `loggit.json`: Log file created each time the application is ran. It contains information about the uploaded packages and application errors. For example, when the user uploads the example csv file `Upload_file_structure.csv` for the first time, a line similar to this one is added to the `loggit.json` file:
 
   ```
-  {"timestamp": "2020-08-28T18:21:29-0400", "log_lvl": "INFO", "log_msg": "Summary of the uploaded file: Upload_file_structure.csv Total Packages: 3 New Packages: 0 Undiscovered Packages: 0 Duplicate Packages: 3"}
+  {"timestamp": "2020-08-28T18:21:29-0400", "log_lvl": "INFO", "log_msg": "Summary of the uploaded file: Upload_file_structure.csv Total Packages: 3 New Packages: 0 Undiscovered Packages: 0 Duplicate Packages: 0"}
   ```
 
   The log functionality is handled by the `loggit` package. The following files write to `loggit.json`:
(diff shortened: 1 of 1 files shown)


MayaGans:
Thanks @marlycormar  😊

Robert-Krajcik (commented):
re-wording one section

- `communityusage_metrics.R` Uses the fields within the table `CommunityUsageMetrics`, create a `Highchart` graph of package downloads and versions over the lifetime of the package. There is also an area to leave comments, which will be added to the`CommunityUsageMetrics` table.

- `maintenance_metrics.R` Uses the table `MaintenanceMetrics` to create the 9 UI badges associated with the risk metric package, and creates an area for leaving comments,
which will be added to the `MaintenanceMetrics` table. 

- `testing_metrics.R` Uses the fields within the table `TestMetrics` to create a `amAngularGauge`. Also includes a comment section which will be added to the `TestMetrics` table.


- `reportpreview.R` generates an overall description of the package, the decision status of the package, and the overall comment of the selected package.  It uses `cum_report.R`, `mm_report.R`, and `testing_metrics.R` to repeat the code seen in the other tabs above and create a singular, downloadable report.<br>
Currently:<br>
    + `cum_report.R` Repeats the code of `communityusage_metrics.R`
    + `mm_report.R`  Repeats the code of `maintenance_metrics.R`
    + `tm_report.R`  Repeats the code of `testing_metrics.R`

MayaGans:
This file is incredibly helpful, thank you for reading and editing! I'll send this visual out via email too so others can edit, but wanted to leave this here as a point of discussion for potential restructuring of the files since this document uncovered some redundancy we might want to consider minimizing: 

Screen Shot 2020-09-21 at 7 46 40 AM

Where each different color in the concept map is a different module (or function). Hopefully this is helpful, and I'd be happy to include this within the document if we adopt a new structure to the UI. Looking forward to getting your thoughts.

Robert-Krajcik:
> This file is incredibly helpful, thank you for reading and editing! I'll send this visual out via email too so others can edit [a very high tech powerpoint file 😆 ], but wanted to leave this here as a point of discussion for potential restructuring of the files since this document uncovered some redundancy we might want to consider minimizing:
> 
> Screen Shot 2020-09-21 at 7 46 40 AM
> 
> Where each different color in the concept map is a different module (or function). Hopefully this is helpful, and I'd be happy to include this within the document if we adopt a new structure to the UI. Looking forward to getting your thoughts.

This looks pretty good @MayaGans 

Now what about `dbupload.R`?
Note that the Maintentance Metrics and Testing metrics are derived from the same source:
`riskmetrics` functions `pkg_ref()` , `pkg_assess()` and `pkg_score()`
From what I can tell the pkg_ref() reads the DESCRIPTION file if the package is installed.
Otherwise it web-scrapes the information from `"https://cran.r-project.org/web/packages/"` (or some other source)
`dbupload.R` web-scrapes by default.
Also note that there is an overall comment loaded by `sidebar.R` and used by `reportpreview.R`
We should probably add that to the comment module

MayaGans:
Hopefully we can talk about that more on the call this afternoon @Robert-Krajcik  since you're way more familiar with that code than I am. I assumed `dbupload.R` just created functions that are used inside the app and we can recycle all those functions to be called within modules? ...Maybe? Curious to hear how you'd structure and include things!

PR #78, opened 2020-11-18
Refactoring maintmetrics

ghost:
This will allow us to report on new metrics as they are developed in the `riskmetric` package.


Robert-Krajcik:
The .csv file created associates the names from `pkg_assess()` and `pkg_score()` with labels for the infoboxes.
Always a better idea to leave this as "easily modifiable data" instead of leaving it in the script.

Now when we get a different metric, all we need to do is update the .csv file

marlycormar:
This is a great starting point for abstracting metrics and creating generic info boxes.

We need to provide a more user-friendly info in some of the boxes. For instance, see the dplyr metrics as the app stands now (left) vs. as it would look like after merging these changes (right).

Metrics for dplyr             |  Metrics for dplyr + PR
:-------------------------:|:-------------------------:
dplyr  |  dplyr

One way to address this issue is to provide further info on the .csv. There are some caveats but we can discuss them in our meeting today.

Modules/dbupload.R
```diff
@@ -163,31 +163,87 @@ metric_mm_tm_Info_upload_to_DB <- function(package_name){
...
+  #             "has_maintainer", "has_news",     "has_source_control" ),
+  #   mm_label = c("News is current?","Presence of vignettes?","Bugs publicly documented?",
+  #                "Bug closure","Documentation","Associated website URL?",
+  #                "Has a maintainer?","Has NEWS?", "Source code public?")
+  # )
+  
+  metrics_to_read <- file.path("Data", "maint_metrics_labels.csv")
+  vartibbl <- readr::read_csv(metrics_to_read, col_types = cols(.default = "c"))
+  
+  # any label that does not have "?" at the end is assumed percentage
```
marlycormar, line 189: The labels may change in the future, so relying on them may not be feasible. An approach to avoid this issue may be to add a new column indicating wether the metric is boolean.

Robert-Krajcik:
I have made some updates and the information looks very similar to the original now.

image

But I no longer have permission to post it, @marlycormar

Robert-Krajcik:
Before and After refactoring_maintmetrics.
image
image

We still need to update the Report_doc.Rmd and Report_html.Rmd

Robert-Krajcik:
It looks like most of the functions in helperScript.js are no longer needed.

Robert-Krajcik:
Replaced by #104

PR #79, opened 2020-11-30
Number of downloads using plotly

### LINKED ISSUE #74: Update x-axis on `Number of Downloads` plot
I’ve found a solution that addresses our need for a line plot that operates more like timeline plot (ie allows for panning, etc). It appears plotly has those functionalities built-in! Please take a look and play with the widget on this page, which includes the code to reproduce. I also like how we can add buttons to the at the top for common timespans (like 1yr, 2yr, etc).
 
https://plotly.com/r/range-slider/

Currently, the plot looks like [this](https://github.com/pharmaR/riskmetric/issues/168#issue-717347723).

AARON-CLARK:
Fixes #74.

PR creates a more interactive plot for users to quickly and easily choose the best x-axis timeline for viewing their package download history. Prior to writing any code, the plot used the highcharter package and looked something like this:

image

This PR's changelog/ to-do list:

- [x] Replace highcharter() with more interactive plotly() object with range slider/ selector layout

- [x] Create `Last Release` button & remove `3 mo` timeline button

- [x] get vertical annotations on version release lines

- [x] Move all time number of downloads into a new info box, not as subtitle on graph

- [x] functionalize/ modularize plot so it can be quickly replicated in cum_report.R, plus both .docx and .html rmarkdown reports

The final plot looks like this:
image

### DIFF
--- a/Reports/Report_doc.Rmd
+++ b/Reports/Report_doc.Rmd
@@ -71,6 +71,8 @@ comment_o <-
    
   time_since_first_release_info <- riskmetrics_cum$time_since_first_release[1] 
   time_since_version_release_info <-riskmetrics_cum$time_since_version_release[1]
+  no_of_downloads_last_year_info <-
+        riskmetrics_cum$no_of_downloads_last_year[1]
   
   riskmetrics_tm <-
     db_fun(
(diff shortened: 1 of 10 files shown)


marlycormar:
The plot looks just like the previous one, even the version numbers! 😎

PR #80, opened 2020-11-30
Clean DB Dashboard's Excel Output

AARON-CLARK:
**Changelog:**

- [x] Removed initially invisible 'Selected' column from `Package Upload History`'s excel download artifact

- [x] cleans up title/ header in the same artifact

- [x] Renames db_dash content from "History of Packages Reviewed" to "Package Upload History" since it is more accurate.

    image


marlycormar:
Looks good @AARON-CLARK!

PR #83, opened 2020-12-07
Various fixes

marlycormar:
### Main changes
- Remove duplicated call to `riskmetric:: pkg_ref` and `riskmetric::pkg_assess`.
- Detect when `export_help` is NA and display this info accordingly (before: the package `binom` for instance displayed "NA%" in green instead of "NA" in grey).
- When uploading a package, show the progress bar with the name of the package being uploaded.
- Remove system sleep time while uploading a package.

### DIFF
--- a/Server/sidebar.R
+++ b/Server/sidebar.R
@@ -7,17 +7,17 @@
 
 
 # Update the sidebar if a decision was previously made.
-observeEvent(input$select_pack, {
+observeEvent(c(input$select_pack, values$selected_pkg), {
   # Suppose package has been selected with a previously made decision.
-  if (nrow(values$selected_pkg) != 0 && values$selected_pkg$decision != "") {
-    # Update the risk slider using the info saved.
-    updateSliderTextInput(
-      session,
-      "decision",
-      choices = c("Low", "Medium", "High"),
-      selected = values$selected_pkg$decision
-    )
-  }
+  req(input$select_pack != "Select")
+  
+  # Update the risk slider using the info saved.
+  updateSliderTextInput(
+    session,
+    "decision",
+    choices = c("Low", "Medium", "High"),
+    selected = values$selected_pkg$decision
+  )
 })
 
 
--- a/Server/sidebar.R
+++ b/Server/sidebar.R
@@ -5,10 +5,8 @@
 # License: MIT License
 #####################################################################################################################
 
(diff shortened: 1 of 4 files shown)


Server/sidebar.R
```diff
@@ -5,29 +5,27 @@
...
-        "decision",
-        choices = c("Low", "Medium", "High"),
-        selected = values$selected_pkg$decision
-      )
-    }
+# Update the sidebar if a decision was previously made.
+observeEvent(input$select_pack, {
+  # Suppose package has been selected with a previously made decision.
+  if (nrow(values$selected_pkg) != 0 && values$selected_pkg$decision != "") {
+    # Update the risk slider using the info saved.
```
AARON-CLARK, line 13: Hi @marlycormar, if you want to change this to an observeEvent, you'll have to add  this event to the observer: `values$selected_pkg` plus a req() statement such as: `req(input$select_pack != "Select")`.

I tested it with only `input$select_pack` and `nrow(values$selected_pkg)` was producing an unexpected result: the values were always lagging behind whatever package I clicked in the UI. To confirm this, just add these print statements to the observeEvent:

`print(paste("values$selected_pkg:", values$selected_pkg))
  print(paste("nrow(values$selected_pkg):", nrow(values$selected_pkg)))`

I'll submit a small PR to this branch if this doesn't make sense. Thanks!

AARON-CLARK (changes requested):
Hi @marlycormar, just one requested change. Besides that, everything else looks good!

Server/sidebar.R
AARON-CLARK: Hi @marlycormar, I tried to push a new branch to your risk_assessment fork but got denied:

image

I may have to be added as a collaborator if applicable. It's not a big issue though, the code is small enough to just paste it here:

image

Hope that helps clarify.

marlycormar: Good catch @AARON-CLARK!

I added lines 11 and 12 to the PR (line numbers from your screenshot). Do you think we still need like 18? The observeEvent should filter out empty inputs. I tested removing that conditional and it seems to work as expected.

AARON-CLARK: It seems to work without line 18!

AARON-CLARK (approved):
LGTM!

PR #91, opened 2020-12-14
Use db_ins() when INSERTing records into COMMENTS table

Robert-Krajcik:
fixes WARNING message

`Warning in result_fetch(res@ptr, n = n) :
  SQL statements must be issued with dbExecute() or dbSendStatement() instead of dbGetQuery() or dbSendQuery().`


AARON-CLARK (approved):
Looks great

PR #92, opened 2020-12-14
Add riskmetric version to reportpreview and reports

Robert-Krajcik:
Addresses #85 The version of riskmetric doest appear when a report is ran

### DIFF
--- a/Server/reportpreview.R
+++ b/Server/reportpreview.R
@@ -18,6 +18,8 @@ output$gen_info <- renderText({
       )
     )
   
+  riskinfo <- package_info("riskmetric", dependencies = FALSE)
+  
   paste(
     "<h2><b>Package:</b> ",
     pkg_GenInfo$package,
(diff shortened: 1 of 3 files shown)


Server/reportpreview.R
```diff
@@ -41,7 +43,8 @@ output$gen_info <- renderText({
     "</h4>",
     "<h4><b>Published:</b>",
     pkg_GenInfo$published,
-    "</h4>"
+    "</h4>",
+    "<h4> Using riskmetric version",riskinfo$loadedversion,"</h4>"
```
marlycormar, line 47: Hi @Robert-Krajcik, it would be best to move this information below as until that point the information displayed is about the package being reviewed.

How about below the overall risk?
https://github.com/pharmaR/risk_assessment/blob/f8e79988d58d8e65db0bbc98676451bcf1bc9a15/Server/reportpreview.R#L57

We can put a label in bold: `**riskmetric version:** [some_version]`.

marlycormar: This PR fixes the problem =) I have only one comment above.

Robert-Krajcik:
I was able to move the `riskmetric version:` line down and increase the font size, so it doesn't look like it's part of the package
image

PR #93, opened 2020-12-15
fix for title not displaying correctly on browser tab

Robert-Krajcik:


app.R
```diff
@@ -34,14 +34,14 @@ options(
...
-        target = "_blank",
-        tags$img(src = "logo.png", class = "logo_1"),
-        tags$img(src = "logo_no_text.png", class = "logo_2")
-      )
+      "R Package Risk Assessment App"
+      # tags$a(
+      #   title = "PharmaR Home Page",
+      #   href = "https://www.pharmar.org/",
+      #   target = "_blank",
+      #   tags$img(src = "logo.png", class = "logo_1"),
```
marlycormar, line 42: Hi @Robert-Krajcik, removing these lines will also remove the logo on the login page and the dashboard.

Robert-Krajcik: I guess I'll try something else then

Robert-Krajcik:
OK this seems to work

marlycormar:
It works Robert!

Issue #53, opened 2020-09-14
Remove code duplication


marlycormar:
There is a LOT of code duplication in the app. We should create functions/modules to address this issue.

Related open issues
https://github.com/pharmaR/risk_assessment/issues/32
https://github.com/pharmaR/risk_assessment/issues/33
https://github.com/pharmaR/risk_assessment/issues/34

Robert-Krajcik:
And it's more pervasive than I thought.
The code is also duplicated in Reports/Report_doc.Rmd and Reports/Report_html.Rmd

Robert-Krajcik:
I have started creating functions for 
comment module, single card module, graph module, and gauge module.

PR #47 move comments handling out of _metrics will be incorporated into another PR

Robert-Krajcik:
Current plan:
Referring to @MayaGans graphic in #58 Add Contributing file

![image](https://user-images.githubusercontent.com/62262182/95111831-1220ad00-070e-11eb-9215-91ce258ef776.png)

I plan to create 5 functions here.  One for the comments, two for the infoBoxes, one for the highchart, one for the angular guage

Then we can have
-  **both** `reportpreview.R` **and** `maintenance_metrics.R` source `mm_report.R`
-  **both** `reportpreview.R` **and** `communityusage_metrics.R` source `cum_report.R`
-  **both** `reportpreview.R` **and** `testing_metrics.R` source `tm_report.R`

We will need to add a bit of code in the mm, cum, and tm report scripts similar to the following:

```{r, eval=FALSE}
out_mm <- reactive({
  req(input$tabs %in% c("mm_tab_value","reportPreview_tab_value"))
  
  if (input$tabs == "reportPreview_tab_value") {
    return(list("vignette1", "website1", "hasnews1",
                "newscurrent1", "bugtrack1", "bugstatus1",
                "exporthelp1", "source_pub1", "pack_maint1",
                "mm_commented1")) }
  else {
    return(list("vignette", "website", "hasnews", 
                "newscurrent", "bugtrack", "bugstatus",
                "exporthelp", "source_pub","pack_maint",
                "mm_commented"))
  }
})

# End of the observe's'

# Start of the render Output's'

# 1. Render Output Info box to show the information on VIGNETTE Content.
observeEvent(input$tabs, {
  req(out_mm())

# output$vignette <- renderInfoBox({
output[[out_mm()[[1]]]] <- renderInfoBox({
```
[...]

PR #100, opened 2021-01-14
Simplify comments

Robert-Krajcik:
Partially addresses #53 Remove code duplication


Robert-Krajcik:
Something doesn't look right.  I'll do another one shortly.

PR #104, opened 2021-01-19
Upd maint metrics

Robert-Krajcik:
This is refactoring maintenance metrics applied to prepare_db_for_weights


Robert-Krajcik:
The metric.csv file is identical to the metric table
and will be replaced once  #105 gets merged.

Robert-Krajcik:
Hi @marlycormar and @AARON-CLARK 

This maint metrics display for package `tidyr`
uses **only** data from riskmetric `pgk_assess()` and/or `pkg_score()`

image

It assumes the table `metric` looks like this:
image

And if more metrics are added at some later time, we can accommodate them easily.
In fact, you would not need to delete/define all the tables, just this one.

```{r, eval=FALSE}
db_ins("drop table metric")
query <- paste("CREATE TABLE IF NOT EXISTS metric(
   id INTEGER PRIMARY KEY AUTOINCREMENT,
   name         CHAR, 
   as_label     CHAR, /* label from pkg_assess() */
   sc_descr     CHAR, /* label from pkg_score()  */
   is_thumb     INT,  /* SQLite does not have a separate Boolean storage class*/
   class        CHAR, /* class = maintenance or test */
   weight       REAL   /* initialized to 1, but could be any number from 0-1 */
);")
db_ins(query)
```

PR #105, opened 2021-01-20
initialize metric table dynamically using pkg_assess()

Robert-Krajcik:
Instead of doing this manually
```{r, eval=FALSE}
INSERT INTO metric (name, description, class, weight) values 
('package_has_vignettes', '', 'maintenance', 1),
('news_is_current', '', 'maintenance', 1),
('has_bug_reports', '', 'maintenance', 1),
('package_has_website', '', 'maintenance', 1),
('news_is_current', '', 'maintenance', 1),
('has_a_package_maintainer', '', 'maintenance', 1),
('source_code_is_public', '', 'maintenance', 1),
('exported_objects_with_documentation', '', 'maintenance', 1),
('status_of_last_30_reported_bugs', '', 'maintenance', 1),
('covr_coverage', '', 'test', 1);
```
This code is built for you dynamically and any new metrics are easily accommodated:
```{r, eval=FALSE}
INSERT INTO metric (name, description, class, weight) values
('news_current','NEWS file contains entry for current version number','maintenance',1),
('has_vignettes','number of discovered vignettes files','maintenance',1),
('has_bug_reports_url','presence of a bug reports url in repository','maintenance',1),
('bugs_status','vector indicating whether BugReports status is closed','maintenance',1),
('export_help','exported objects have documentation','maintenance',1),
('has_website','a vector of associated website urls','maintenance',1),
('has_maintainer','a vector of associated maintainers','maintenance',1),
('has_news','number of discovered NEWS files','maintenance',1),
('has_source_control','a vector of associated source control urls','maintenance',1),
('covr_coverage','Package unit test coverage','test',1)
```

PR #106, opened 2021-01-22
Replace queries to the MaintenanceMetric table with queries to package, metric, package_metrics

marlycormar:


Robert-Krajcik:
Hi @marlycormar 
How about if we do #104 and #105 first, and then look at what else we need to change?
The metric table is going to be initialized dynamically; no need to fix it.

Robert-Krajcik (approved):
Well, it works. @marlycormar 

But this is still some hardcoding here.
I was looking at refactoring all this so we don't have to touch the code when we get new metrics.
But right now my efforts feels like I'm trying to fix the landing gear while we're flying the plane.

Robert-Krajcik:
As I think about it,
The WEIGHT on the metric and package_metrics tables should be REAL instead of INT

marlycormar:
> Well, it works. @marlycormar
> 
> But this is still some hardcoding here.
> I was looking at refactoring all this so we don't have to touch the code when we get new metrics.
> But right now my efforts feels like I'm trying to fix the landing gear while we're flying the plane.

Hey @Robert-Krajcik, thanks for reviewing the PR!

It makes sense that there is still hardcoding in the PR as this PR is not meant to fix that issue in the app. This PR prepares the db for the addition of weights, and although I did have to refactor some of it on the way, it was not the intended focus.

marlycormar:
> As I think about it,
> The WEIGHT on the metric and package_metrics tables should be REAL instead of INT

Not sure why the type INT worked, but I changed it to DECIMAL as I found it recommended over REAL. What do you think?

Robert-Krajcik:
I know REAL is equivalent to R dbl (double-precision floating point)
SQLite recommended DECIMAL over REAL?  
How many decimal points can it handle?
If you think it will work, then go with it.

marlycormar:
To be honest, I searched for SQL and not SQLITE. Searching now for SQLITE, it seems that it 'understands' DECIMAL types but it is not native to SQLITE. Although it works as it is, I don't like that it is not native to SQLITE. So I will update the types.

marlycormar:
What is puzzling is that it was working before with INT ¯\_(ツ)_/¯

marlycormar:
Done @Robert-Krajcik!

PR #110, opened 2021-03-22
Drop tm panel

Robert-Krajcik:
Drop tm panel from dashboard_screen
and remove testing metrics from reportpreview as well as the docx and html reports


AARON-CLARK (approved):
Looks great @Robert-Krajcik

PR #120, opened 2021-04-27
Add ui elements to view/change metrics weights

marlycormar:
Created a new section on the database view (database icon page) that contains a table and two dropdown inputs to update the weights. This new section is in a new uiOutput so that we can later show it only for authorized users.

Based on todays meeting, these are some of the items we discussed:
- [ ] Is the UI easy to understand? Any improvements in this direction?
- [ ] We should add documentation on how the weights are calculated. After you guys review it, we can discuss what were your main concerns and have them into account when constructing the documentation.
- [ ] Should we (can we?) move this UI to the shinymanagers admin view?
- [ ] The weights of each metric should be displayed on the tabs and the final report (and in a table on the db view?).

### DIFF
--- a/Data/maintenance.txt
+++ b/Data/maintenance.txt
@@ -1,1 +1,0 @@
-Best practices in software development and maintenance can significantly reduce the potential for bugs / errors.  Package maintainers are not obliged to share their practices (and rarely do), however the open source community provides several ways of measuring best practice.  The R Validation Hub proposes the following metrics.
--- a/Modules/dbupload.R
+++ b/Modules/dbupload.R
@@ -167,7 +167,7 @@ metric_mm_tm_Info_upload_to_DB <- function(package_name){
   
   # Insert all the metrics (columns of class "pkg_score") into the db.
   # TODO: Are pkg_score and pkg_metric_error mutually exclusive?
-  for(row in 1:length(metric_weights_df)){
+  for(row in 1:nrow(metric_weights_df)){
     metric <- metric_weights_df %>% slice(row)
     # If the metric is not part of the assessment, then skip iteration.
     if(!(metric$name %in% colnames(riskmetric_score))) next
--- a/Modules/dbupload.R
+++ b/Modules/dbupload.R
@@ -148,7 +148,7 @@ metric_mm_tm_Info_upload_to_DB <- function(package_name){
     pkg_assess()
   
   # Get the metrics weights to be used during pkg_score.
-  metric_weights_df <- db_fun(paste0("SELECT name, weight FROM metric"))
+  metric_weights_df <- db_fun(paste0("SELECT id, name, weight FROM metric"))
   metric_weights <- metric_weights_df$weight
   names(metric_weights) <- metric_weights_df$name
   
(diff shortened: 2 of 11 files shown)


Robert-Krajcik:
Just temporarily modifying this section of code to db_dash_screen.R so I can complete some testing...
The metric scores are dynamically updated this way.

```{r, eval=FALSE}
# Save new weight into db.
observeEvent(input$update_weight, {
  validate(
    need(is.numeric(input$metric_weight), "Please select a valid numeric weight.")
  )
  update_metric_weight(input$metric_name, input$metric_weight)
  print(paste("updating metric weight for",input$select_pack))
  db_ins(paste0("delete from package_metrics where package_id = ", 
                "(select id from package where name = ","'", input$select_pack, "')") )
  metric_mm_tm_Info_upload_to_DB(input$select_pack)
  
  values$db_pkg_overview <- update_db_dash() %>%
    mutate(last_comment = as.character(as_datetime(last_comment))) %>%
    mutate(last_comment = ifelse(is.na(last_comment), "-", last_comment)) %>%
    mutate(decision = ifelse(decision != "", paste(decision, "Risk"), "-")) %>%
    mutate(was_decision_made = ifelse(decision != "-", TRUE, FALSE)) %>%
    select(name, version, score, was_decision_made, decision, last_comment)
  
},ignoreInit = TRUE)
```

Robert-Krajcik:
I think the UI you have is easy enough to understand.
But I also think we can and should update the score for the selected package as soon as the weight or weights are updated.
I know this will cut across all packages but it would be good to give the shinymanager some immediate feedback.

I am not sure we need to insert this UI into the shinymanagers admin view.
We can find out who is or isn't an admin using some reactive variables provided.
But I'll go back to the other PR and test it out.

Robert-Krajcik (changes requested):
The UI seems easy enough to understand.
Maybe adding documentation on how the weights are calculated can be a separate PR.
But I think we can and should update the score for the selected package as soon as the weight or weights are updated.
The weights of each metric should be displayed on the tabs and the final report -- maybe that can be another PR
I don't think we need to move this UI to the shinymanagers admin view.  At least not now.
Too many interdependencies, shinymanager still needs some work.

Robert-Krajcik:
Another suggested change, @marlycormar 
in dbupload.R
update package_metrics with the metric_id$weight instead of "1"
look at lines 173 and 197
This will update package_metrics only for the selected package.

marlycormar:
@Robert-Krajcik, @AARON-CLARK 
What do guys think if in a separate PR we expand the 'change of weights ui' to include another button that updates the risk of each package based on the new weights (instead of automatically updating the risk of each package). Since updating the risk of each package takes time, we can take advantage of such button to add a shiny waiter as well warn the user before proceeding. Also, if the user makes a mistake while updating a weight, or if he wants to update multiple weights before recalculating the risk, then he has the chance to do so.

Robert-Krajcik:
That would work, @marlycormar 
We could also add an overall comment and/or loggit for the selected package 
And that would give the admin a chance to update multiple weights first., as you mentioned.

Robert-Krajcik:
Having an issue with the maintenance metrics reporting, @marlycormar 
image

It's only showing has_bug_reports_url, which is row #3 in metric_weights_df
In dbupload.R, I think we want:
 `for(row in 1:row(metric_weights_df))`
not
  `for(row in length(metric_weights_df))`

Robert-Krajcik:
If we are going to update the weights in metric db,  @marlycormar 
let's update the weights in package_metrics for that package at the same time.
```{r, eval=FALSE}
# Save new weight into db.
observeEvent(input$update_weight, {
  validate(
    need(is.numeric(input$metric_weight), "Please select a valid numeric weight.")
  )
  
  update_metric_weight(input$metric_name, input$metric_weight)
  update_pkg_metricswt(input$metric_name, input$metric_weight, input$select_pack)
})
```
Function to be added to utils.R
```{r, eval=FALSE}
update_pkg_metricswt <- function(metric_name, metric_weight, select_pack) {
  db_ins(paste0(
    "UPDATE package_metrics ",
    "SET weight = ", metric_weight, " ",
    "WHERE metric_id = (select id from metric where name = ","'", metric_name, "') and ",
    "package_id = select(id from package where name =", "'", select_pack, "')"
  ))
}
```

Robert-Krajcik:
Updated the function, putting all SQL code in uppercase (and changing "select(id"  to "select(id" )
```{r, eval=FALSE}
update_pkg_metricswt <- function(metric_name, metric_weight, select_pack) {
  db_ins(paste0(
    "UPDATE package_metrics ",
    "SET weight = ", metric_weight, " ",
    "WHERE metric_id = (SELECT id FROM metric WHERE name = ","'", metric_name, "') AND ",
    "package_id = (SELECT id FROM package WHERE name =", "'", select_pack, "')"
  ))
}
```

Robert-Krajcik (approved):
All working now.

Issue #124, opened 2021-05-11
Update risk and metric weights per package by admin request


marlycormar:
- [x] Add a button that allows admins to update the risk of all packages in the db as well as the package weights after updating the weight of each metric.
- [x] If such button is pressed, the following items should be addressed:
   - [x] If final decision was made in a package, restart it (i.e., allow the user to submit another final decision).
   - [x] If a package contains comments or a final decision was made on it, then add a comment on every tab saying how the risk and weights changed, and that the comments, final decision may no longer be applicable.
   - [x] Write to the log file stating that the weights and risk have changed.
   - [x] Ask the admin if he/she wants to download a snapshot of the db before updating every risk (re-use the functionality implemented in #123).
   - [x] Add a pop message warning the admin of the permanent change he/she will make.

PR #119, opened 2021-04-20
shinymanager secure database implementation

Robert-Krajcik:
ref: [shinymanager web page](https://datastorm-open.github.io/shinymanager/index.html)
created a separate database for credentials.
I can at least say that it works.
If you sign on as an admin, you get access to admin functions (lower right corner)
image
image

PR #125, opened 2021-05-14
update risk and package weight by admin

### LINKED ISSUE #123: Allow admins to download a snapshot of the DB

Cytel-Inc:
closes #123, #124, and #134.


Robert-Krajcik:
This PR also addresses issue #123 

There is a button to download a snapshot of the package table as a .csv file
And a button to update the risk and package weights
A modal appears when this button is pressed.
image

If the submit button is pressed,
An overall comment is added to every package that had a comment added to it.
Any final decisions are reverted.
The package weights and risk scores are updated for every package.
An info message is written to the log file stating that the weights and risks have changed.
image

Robert-Krajcik:
When you have some time, can you take a look at what I have here and give me some comments on it, @marlycormar ?

Robert-Krajcik:
Updated Modal Dialog message:
image

marlycormar:
Hey @Robert-Krajcik, I have a few more comments but we can discuss them tomorrow.

Robert-Krajcik:
@AARON-CLARK 
I'm going to let  you take over on this one as I'm going to be OOO the next three days.

marlycormar:
Based on our discussion today:
- [ ] Fix bug: the app fails when updating the risk with an empty list of packages in the db.
- [ ] Add a how-to info on the View/change weights panel (e.g. an overall vignette, button-wise info, etc.).
- [ ] Change the package snapshot download to database download (maybe wiser to dump the db).

marlycormar:
Hi @Robert-Krajcik, I checked the db snapshot download, and it seems to work!

I'll check the weights update next.

marlycormar:
Can we download the `dbdatabse.sqlite` file?

marlycormar:
Hi @Robert-Krajcik, I checked the risk update process and it updates the risk of each package on the db and resets final decisions as we discussed. The only suggestion I have is to add a comment on each tab instead of on the overall comment on the side bar (since the latter is too constricted for this text).

Robert-Krajcik:
OK, I got the second request,
I'm still looking into how to use `downloadHandler()` to download the `dbdatabse.sqlite` file

AARON-CLARK:
Hi @marlycormar, when you get time, can you review this PR?

marlycormar:
Hi @Robert-Krajcik, thank you for all these changes! I think we are getting there.

A few comments:

1. Regarding downloading the `database.sqlite`, I made a few tweaks to your code (but it is mainly your code) and I was able to get it to download. See this [reproducible script](https://github.com/pharmaR/risk_assessment/files/7565118/testing_sqlite_download.R.zip).

2. Regarding updating the database and adding comments reflecting this update, I still see a comment on the sidebar and now a new one on the reports tab. I think we should add a comment in both the Maintenance Metrics and the Community Usage Metrics, and remove the rest (@AARON-CLARK @Robert-Krajcik do you guys agree?).

AARON-CLARK:
Hi team, couple of items, some of which are comments and others are issues:

1. First, is it possible to arrange these "Update Weights" UI elements to a new position on top of the table since it only updates the table and not the database? The other two buttons perform a separate function in my mind, regarding the database. And instead of saying "Update Weights" maybe we could re-word to say "Change weights"? Only because "update" may be imply we are updating risk scores (prematurely).
image

1. I think it'd make sense to display the database downloader button next, since that's the recommended workflow, right? Backup the database before changing it? Also, when clicking this button, I don't think you need to show the temp file location (which is difficult to read) since the file downloads right in the browser. Maybe just say "database.sqlite has been downloaded to your browser." and leave it at that?
image

1. Next, I think the wording "final decisions will be reversed" sounds misleading because that suggests we're changing the decision. For example, if we said a package was "low risk", to reverse it would mean we changed the package decision to "high risk". Maybe it'd be best to say "Final decisions on packages will be removed" or "dropped". While we're here, could we join the first two bullets into one for simplicity: "The package weights will be applied and risk metric scores re-calculated". For the last bullet, can we say just "Comments" and not "Overall comments"? Aren't we adding comment everywhere? @marlycormar, I don't think we should remove comments from anywhere. Sorry if I'm out of the loop/ misunderstood. 
image

1. So, I made a final decision for `haven` and then tried to change the "has_news" weight before re-calculating each pkg's risk and the app crashed. Whomp whomp!
image

Robert-Krajcik:
Now it looks like this
image

Robert-Krajcik:
I selected package `glue` and added comments and an overall comment and final decision.
Now after pushing `re-calculate` button:
The overall comment is gone and a new comment is added to the maint metrics and cum metrics tabs.
As stated in the `Confirm Decision` modal,
Comments will be added for each package that had a comment

image

Robert-Krajcik:
updated Confirm Decision modal
image

Robert-Krajcik:
Ready for another review, @AARON-CLARK and @marlycormar

AARON-CLARK:
Hi @Robert-Krajcik, thanks for making all those changes! I think it looks a lot better. I did another review and I have another set up comments and (unfortunately) issues that popped up.

First, if the package score improves after re-calculation, do you think we could retain the original decision but still leave a comment? I don't think anyone would want to re-evaluate a package if it's risk score got even better!
Second, people probably won't understand what this comment stuff means. Can you change the wording to "The risk re-calculation will be logged as a comment for each package that had a comment or final decision." Just want to check, if a package doesn't have a comment, but it has a final decision, does the re-calc auto comment get added?
image



Change text to "Applying weights & updating risk scores"
image

Issue: all final decisions were not dropped after recalculating weights! Before re-calucation:
image

After re-calculation. For some reason, dplyr's decision was kept... strange!
image

Regarding the admin's auto-generated re-calculation comment, wondering if we can make it more informative of what actually happened? Could we say something like, "Risk score re-calculated based on new metric weighting. New risk score = .07 (old score = .09). Final decision may no longer be applicable." I think knowing that the risk score went up or down is important to know. For example, if it went down, you probably would want to keep the final decision (should we consider doing that automatically?) but if it went up, you'd probably want to re-review the package. I don't think we need to include what weights changed since (1) that could be tedious to put into a comment, (2) if they really want to know, they can look in the database history, and (3) weights are only visible to the admin to anyway... so other users probably shouldn't be privvy to how the admin changed the weights. Thoughts?
image

Robert-Krajcik:
if a package doesn't have a comment, but it has a final decision, does the re-calc auto comment get added?
No, it does not at this time.

Robert-Krajcik:
Thanks for all your feedback, @AARON-CLARK 
I will make the changes you wanted and fix the final decisions part.

Robert-Krajcik:
For "either had a comment or a decision"
```{r,eval=FALSE}
  cmts_db <- unique(bind_rows(db_fun("select distinct comm_id as package_name from Comments"),
                              db_fun("select distinct name as package_name from package where decision != ''")))
```

Robert-Krajcik:
Regarding the admin's auto-generated re-calculation comment
I think the new scores should be next to the original score in the uploaded packages section
that is only visible to the admin.

AARON-CLARK:
> I think the new scores should be next to the original score the uploaded packages section that is only visible to the admin.

Image below for reference. So, you would have a "Score" column, and then an "Last Score" and "Last Score Datetime" columns added? I think it'd be easier for historical purposes to scroll through a list of comments to see the changes. I thought the "Uploaded Packages" section was visible to all users and the "View / Change Weights" section was only visible to admins?  Also, just noticed that the title "Select metric" on the change weights datatable doesn't really make sense since you select metrics using the dropdown list above the table. Could we just title it "Current Risk Score Weights by Metric"?

image

Robert-Krajcik:
Yeah I guess you are right about the scores.
I know in the package db we have a `weighted_score` variable that we aren't using.

I have an idea, @AARON-CLARK 
How about if I had this PR over to you?

AARON-CLARK:
> How about if I had this PR over to you?

Hi @Robert-Krajcik, 
Sure, I can help write some code, but not till I'm back in office Nov 29. If you and @marlycormar are around feel it's ready to merge Mon - Wed next week, please do! Happy Thanksgiving!

Robert-Krajcik:
I have been working on these risk assessment PRs all day,  I definitely need a break from this one.
Happy Thanksgiving, @AARON-CLARK

AARON-CLARK:
- [ ] Do not save weights immediately to database, before clicking re-calc button since they are still temporary. That way, we can display old vs. new.
- [ ] disable re-calculate button if no weights have changed
- [ ] adjust automated comment to include old risk score and new score so everyone knows if the score went up or down.
- [ ] Center the values of the wts table
- [ ] Add option to retain decisions if scores improve (go down)?

Robert-Krajcik:
Just wondering if we should update the maint metrics and community usage comments for all packages,
or just those that had a comment, which is what we are doing now?
`The risk re-calculation will be logged as a comment for each package that had a comment or final decision.`

AARON-CLARK:
> Just wondering if we should update the maint metrics and community usage comments for all packages,
> or just those that had a comment, which is what we are doing now?

@Robert-Krajcik, hmmm. I think I like what we are doing now, but I could be influenced to do it across the board (for all pkgs in the db) too. Our original logic was that if there is no decision nor comment, then maybe no one's even started reviewing the package yet, so why create a comment? But on the flip side, there is also no harm in inserting a comment either. It certainly would be easier / more consistent to create those automated comments for all pkgs. @marlycormar, what do you think?

marlycormar:
> > Just wondering if we should update the maint metrics and community usage comments for all packages,
> > or just those that had a comment, which is what we are doing now?
> 
> @Robert-Krajcik, hmmm. I think I like what we are doing now, but I could be influenced to do it across the board (for all pkgs in the db) too. Our original logic was that if there is no decision nor comment, then maybe no one's even started reviewing the package yet, so why create a comment? But on the flip side, there is also no harm in inserting a comment either. It certainly would be easier / more consistent to create those automated comments for all pkgs. @marlycormar, what do you think?

I also think is a good idea to leave a comment on all the packages, even those that don't have any comments/final decision. As @AARON-CLARK said, it is good for consistency. In addition, if a reviewer quickly loads a package into the app to see its risk, it is happy with it but doesn't leave a comment, and then returns to the app after weights have been updated, he will not understand why the risk is different.

AARON-CLARK:
> and then returns to the app after weights have been updated, he will not understand why the risk is different.

Good point @marlycormar.

I've updated this branch to add notification comments to all packages when risk is recalculated. But I did make it "smart" to adjust the comment when a final decision or overall comment was dropped. Below, the bottom comment was added when no decision was made. And the top comment was generated when a decision was made. Thanks for the feedback. This branch is ready for a final review @Robert-Krajcik or @marlycormar!

image

Robert-Krajcik (approved):
I love  your changes, @AARON-CLARK 
This looks great to me.

AARON-CLARK:
Hi @marlycormar, did you also want to review this PR? Just checking so we don't merge pre-maturely.

marlycormar:
> Hi @marlycormar, did you also want to review this PR? Just checking so we don't merge pre-maturely.

Hi @AARON-CLARK, thanks for checking! Yes, I would like to give it a quick look. Doing that now...

AARON-CLARK:
I see there are a couple merge conflicts now that #119 went through to master. Stay tuned and I'll update.

AARON-CLARK:
Hi @marlycormar, since you viewed the PR earlier today, I made the following changes:

- [ ] fixed merge conflicts
- [ ] removed pagination on DT weights table
- [ ] big fix in "assessment criteria" modal on 'Risk Calc' since it was pulling
- [ ] removed 'time since first release' and 'time since version release' from "assessment criteria" modal since they aren't assessed in `riskmetric`
- [ ] Tried to make the purpose of 'new weights' column more obvious, as seen below

image

AARON-CLARK:
Per our discussion yesterday (12/6), I've hardcoded the `covr_coverage` metric to have a weight of zero. The `metrics` table will initialize to 0. In addition, I've ensured the user cannot change the value of the `numericInput()` away from zero, and even display a little warning/ note below the table when `covr_coverage` is selected. I checked the current `riskmetric` list of [github issues](https://github.com/pharmaR/riskmetric/issues) and didn't see anything regarding `NA` values for `covr_coverage`. Is it even on their radar? I'd love to include a hyperlink to an open issue in my note. See below:

image

Last, I added the same note to the `assessment_criteria` modal:

image

I think that covers all our bases for now! Please let me know your thoughts @Robert-Krajcik & @marlycormar.

Robert-Krajcik:
Two questions remain: 
- Do any of these metrics: r_cmd_check, remote_checks, and exported_namespace have any influence on the risk score?
- Should we be tracking them (creating infoboxes)?

AARON-CLARK:
> ... have any influence on the risk score?

Yes, I think Marly demonstrated that all the metrics have an influence on the risk score. If you want to double check, you could use the script below. Feel free to check different packages. Also, my version of `riskmetric` doesn't have an assessment called "exported_namespace" so I left that out.

```
library(magrittr)
library(riskmetric)

my_pkg_ref <- pkg_ref("editData") 

my_pkg_ref %>%
  pkg_assess() %>%
  pkg_score() %>%
  summarize_scores()

my_pkg_ref %>%
  pkg_assess(assessments = all_assessments()[
    names(all_assessments()) != 'assess_covr_coverage']) %>%
  pkg_score() %>%
  summarize_scores()

my_pkg_ref %>%
  pkg_assess(assessments = all_assessments()[
    names(all_assessments()) != 'assess_r_cmd_check']) %>%
  pkg_score() %>%
  summarize_scores()

my_pkg_ref %>%
  pkg_assess(assessments = all_assessments()[
    names(all_assessments()) != 'assess_remote_checks']) %>%
  pkg_score() %>%
  summarize_scores()
```

They probably should have info boxes, but since that is a big change, it should probably be implemented in a different PR. Perhaps the PR that solves issue #142.

Robert-Krajcik:
I recall that PR#104 was an early attempt at getting away from "hardcoding" an infobox for each metric

AARON-CLARK:
per @marlycormar's request, I'm merging!

PR #131, opened 2021-07-20
Improve database feedback

Robert-Krajcik:
A PR for @marlycormar to test

### DIFF
--- a/Utils/utils.R
+++ b/Utils/utils.R
@@ -84,7 +84,8 @@ db_fun <- function(query){
   
   dat <- dbFetch(rs)
   dbClearResult(rs)
-  if (nrow(dat) == 0 & str_detect(strng, "'Select'") == FALSE) {
+  if (nrow(dat) == 0 
+     & (str_detect(query, "name = 'Select'") == FALSE) && str_detect(query, "name = ''") == FALSE) {
     message(paste0("No rows were returned from db_fun query\n",query))
   }
   dbDisconnect(con)
--- a/Utils/utils.R
+++ b/Utils/utils.R
@@ -42,60 +42,66 @@ create_db <- function(){
   sapply(queries, function(x){
     
     tryCatch({
-      res <- dbSendStatement(
-        con,
-        paste(scan(x, sep = "\n", what = "character"), collapse = ""))
+      rs <- dbSendQuery(con, query)
     }, error = function(err) {
-      # If there is an error, then log it, close the connection, inform the
-      # user, and leave method.
-      loggit("ERROR", paste("create_db error: ", err))
+      message <- paste("dbSendQuery returned",err)
+      message(message, .loggit = FALSE)
+      loggit("ERROR", message)
       dbDisconnect(con)
-      # TODO: show pop up message.
-      stop(err)
     })
-
-    dbClearResult(res)
+    
+    dbClearResult(rs)
   })
   
   dbDisconnect(con)
 }
 
 db_fun <- function(query){
+  errFlag <- FALSE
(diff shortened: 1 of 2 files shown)


Utils/utils.R
```diff
@@ -40,31 +40,69 @@ create_db <- function(){
...
+    tryCatch({
+      rs <- dbSendQuery(con, query)
+    }, error = function(err) {
+      message <- paste("dbSendQuery returned",err)
+      message(message, .loggit = FALSE)
+      loggit("ERROR", message)
+      dbDisconnect(con)
+    })
+    
+    dbClearResult(rs)
```
marlycormar, line 53: Hi @Robert-Krajcik, when I run the app after removing the `database.sqlite` and the `loggit.json` files, I get the following error:

`dbSendQuery returned Error in dbSendQuery(con, query): object 'query' not found`

followed by 

` Error in dbClearResult(rs) : object 'rs' not found `

Robert-Krajcik:
Needed to use `dbSendStatement` instead of `dbSendQuery` and use x instead of query in function.
Also issue message "No rows were returned from db_fun query" if and only if the package name is not 'Select'

Utils/utils.R
Robert-Krajcik: OK I believe I have this figured out, @marlycormar

marlycormar:
Ready for my review.

Utils/utils.R
```diff
@@ -40,31 +40,73 @@ create_db <- function(){
...
+      dbDisconnect(con)
+      errFlag <<- TRUE
+    },
+    finally ={
+      if (errFlag) return(NULL) 
+    })
+  
+  dat <- dbFetch(rs)
+  dbClearResult(rs)
+  if (nrow(dat) == 0 & str_detect(strng, "'Select'") == FALSE) {
```
marlycormar, line 87: Hi @Robert-Krajcik , I get the following error when running the app on this branch:

`Warning: Error in str_detect: object 'strng' not found`

Robert-Krajcik: OK I'll take a look in a few minutes

Robert-Krajcik:
Hi @marlycormar 
Try it now.
The "strng" should have been "query"

marlycormar:
Hey @Robert-Krajcik! Could you share a summary of this PR?

Robert-Krajcik:
Sure, @marlycormar 
I'm planning on spending most of the day tomorrow on risk_assessment

Robert-Krajcik:
The purpose of this PR is to replace un-helpful database warning and error messages
with more descriptive messages using `tryCatch()`

for example, a message text is created
`message <- paste0("db_fun warning:\n",query,"\nresulted in\n",warn)`
where this message shows the actual query that caused the warning.
This gets logged to `loggit.json`

error messages are handled in the same way
`message <- paste0("db_fun error:\n",query,"\nresulted in\n",err)`

For queries  using the `db_fun()` function
an informational message will be issued when no rows were returned by the function. 

For delete, update or insert commands using the `db_ins()` function
a descriptive error message will be displayed,
as well as the message
`zero rows were affected by db_ins() function.`
if no rows were affected.

marlycormar:
I was able to test the db error (by deleting the db while the app was running) but not the db warning. Do you have any tips @Robert-Krajcik?

Robert-Krajcik:
Let me search for something that generates a warning

Robert-Krajcik:
Is the message
`No rows were returned from db_fun query`
helpful at all?
I'm thinking about commenting it out/removing it

PR #144, opened 2021-11-29
Use default `skipLabel` for `rintrojs` guide

### LINKED ISSUE #143: `rintrojs` help box says "Close" instead of "X"
Hi, wondering everyone's thoughts on the following: The `rintrojs` help modal that shifts across the screens has a large bold "Close" word that serves as a button to close the modal. I would argue that the traditional "x" would be more appropriate since it is the industry standard for closing windows. It's not obvious that "Close" is even a button... kind of like a norman door. It's easy to switch, just have to remove [this ](https://github.com/pharmaR/risk_assessment/blob/34fafe42e400c921ebf2badc8548a3f68752fd58/Server/uploadpackage.R#L48) line of code. I'd be happy to open a PR, but don't want to do so if anyone has strong opinions about it?

image

image

AARON-CLARK:
One line of code to close #143. Either of you can review and merge when ready.

### DIFF
--- a/Server/uploadpackage.R
+++ b/Server/uploadpackage.R
@@ -44,8 +44,7 @@ observeEvent(input$help,
           union(upload_pkg_steps()) %>%
           union(sidebar_steps),
         "nextLabel" = "Next",
-        "prevLabel" = "Previous",
-        "skipLabel" = "Close"
+        "prevLabel" = "Previous"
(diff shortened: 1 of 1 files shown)


Robert-Krajcik (approved):
This looks fine.  Tested every tab.

Issue #142, opened 2021-11-09
Sync app metrics with the ones from `riskmetric`


marlycormar:

AARON-CLARK:
I think @Robert-Krajcik started this effort with #105 back in Jan 2020! It'd be worthwhile to review that code before attempting to re-invent the wheel.

As it's highly related, I'd like to also note here that the "Assessment Criteria" modal is pulling it's metric-related data like "Reason for inclusion" from a csv file, as seen [here ](https://github.com/pharmaR/risk_assessment/blob/e5eeb287bed1b34d21eb7595618d4ed4976d5c79/Server/assessment_criteria.R#L69)(with image below) which isn't ideal.

![image](https://user-images.githubusercontent.com/17878953/144666625-a6e69e8b-3216-437f-8b68-a90c86e993ab.png)

![image](https://user-images.githubusercontent.com/17878953/144666921-db0c8ac5-4d48-4435-800c-4ba4287fa14f.png)

Robert-Krajcik:
The labels from `pkg_score` (descr2) help determine if you are looking at a `binary indicator` (thumbs up or down)
or `The fraction of` (a proportion) which can be converted to a percentage.

marlycormar:
Since the metrics are different by version of riskmetric, what do you think if we use: `release-v0.1.1`?. That would be
```devtools::install_github("pharmaR/riskmetric", ref = "release-v0.1.1")```

marlycormar:
These are the metrics of v0.1.1:

1. "assess_news_current"       
2. "assess_has_vignettes"      
3. "assess_has_bug_reports_url"
4. "assess_last_30_bugs_status"
5. "assess_license"            
6. "assess_export_help"        
7. "assess_downloads_1yr"      
8. "assess_has_website"        
9. **"assess_r_cmd_check"**
10. **"assess_remote_checks"**     
11. "assess_has_maintainer"     
12. "assess_has_news"           
13. "assess_has_source_control" 
14. **"assess_covr_coverage"**


In the app, we are missing the metrics in bold. However, they seems to produce only NAs. We can go with @AARON-CLARK's suggestion to put zero weight on these metrics, instead of removing them.

AARON-CLARK:
> These are the metrics of v0.1.1:

Glad we are all on the same version, observing the same metrics and metric scores! Great work!

[...]

PR #151, opened 2021-12-13
Sync metrics and minor tweaks

marlycormar:
- Set `riskmetric` version to **release-v0.1.1**.
- Add metrics **r_cmd_check** and **remote_checks** to match `riskmetric`. These metrics have been added with weight 0.

Related to #142.


global.R
```diff
@@ -42,11 +42,9 @@ package.check <- lapply(
   }
 )
 
-# Install riskmetric from github and load it
-if(!require(riskmetric)){
-  devtools::install_github("pharmaR/riskmetric")
-  library(riskmetric)
-}
+# Install a specific version of riskmetric from GitHub and load it.
+devtools::install_github("pharmaR/riskmetric", ref = "release-v0.1.1")
```
AARON-CLARK, line 46: So, this line of code will re-install the `0.1.1` version every time you run the app. Is that the intended behavior? How about we use an if-then statement using packageVersion("riskmetric") to see if the user has the correct version. We can also add this version to the `DESCRIPTION` file.

image

AARON-CLARK (changes requested):
Hi @marlycormar, comments below! Plus I'll push a commit to import this version of `riskmetric`. Stay tuned.

global.R
marlycormar: Change this line by `library(riskmetric)`. The package version of the app will take care of package versioning.

marlycormar: Hi @AARON-CLARK, I am having trouble pushing the updated `global.R file`. My local git doesn't know how to reconcile my changes with the branches that are now part of this PR. 

Could you please undo the changes I made? Basically remove lines 46-47 and add back 45-49?

AARON-CLARK: done!

AARON-CLARK (approved):
LGTM!

AARON-CLARK:
Hey @marlycormar, this branch looking good. Did you have any other changes planned or can I merge? I'd like to start "packagizing" the app after these changes are pushed through!

marlycormar:
Yes, please.

PR #157, opened 2022-01-12
Add `DESCRIPTION`

AARON-CLARK:
Hey all, I added a `DESCRIPTION` file in this PR, plus some of the `usethis` functions I ran to modify it in a new R script called `001_start.R`, mimicking `golem`. In a consecutive PR, I actually hope to turn this PR into a 'golemized' app-pkg, so I thought it'd be wise to start following their conventions now.

Also, we are currently loading the `tidyverse` which is a "no no" according to `usethis`:

image

So, right now, this PR imports all `tidyverse` packages, but please comment on this thread to tell me which ones we don't need from the following list and I'll remove them.

image


Robert-Krajcik:
Hi @AARON-CLARK 
I know we use dplyr ("mutate") and lubridate and stringr ("str_replace_all")
and rvest ("html_nodes")
and tibble ("as_tibble")
and readr ("read_file")
and purr (map(rev))
probably rlang

Why not start with this, and see if we missed anything down the road

marlycormar:
I played around with the app, I found similar results to Robert.

- dplyr
- rvest ("html_nodes")
- ~~tibble (not needed since "as_tibble" is included in dplyr)~~
- readr
- purr
- rlang (needed for "is_empty")
- stringr


The following packages are part of tidyverse and included in the global.R file, but I didn't find where they were used on the app:

- lubridate
- xml2
- httr

I don't think we are using the following packages:
- ~~ggplot2~~
- ~~tidyr~~

AARON-CLARK:
@marlycormar & @Robert-Krajcik, I updated the `DESCRIPTION` file accordingly. This is good enough to get us started. Later, as we convert the app into a package, we'll be able to more easily identify which packages the `riskassessment` depends on and update the `DESCRIPTION` accordingly.

Issue #158, opened 2022-01-15
Issue with report generation


marlycormar:
When generating reports (either as an html or docx), the loggit wording appears, e.g., 
![Screen Shot 2022-01-16 at 4 31 47 PM](https://user-images.githubusercontent.com/8559654/149678886-6ceb87d7-86c6-44ed-ba4c-c69b97092ca3.png)

PR #159, opened 2022-01-16
Fix reports

marlycormar:
Closes issue #158.


AARON-CLARK (approved):
Hi @marlycormar, great catch finding this issue! I reviewed your PR and I pushed a couple commits to your branch.

First, I merged your branch with `master` and added the `glue` package to the newly added `DESCRIPTION` file.

When downloading reports from the `Report Preview` tab, everything worked great! However, when downloading the zip file of reports from the `Database Overview`, I get the following errors for either html or docx output:

image

image

image

So I adjusted the `params` being used for that second `rmarkdown::render()`  call to fix the issue. 

Thanks!

marlycormar:
Ditto! Thanks for fixing the issue with multiple downloads.

AARON-CLARK:
I just buttoned up some loose ends, but I'm done now. IMO, this PR is ready to merge with `master`.

Issue #163, opened 2022-01-21
`metric_cum_Info_upload_to_DB()` is broken


AARON-CLARK:
Hi all,

The Community Usage plotly is missing pertinent info. I pulled up `dplyr` and compared it against their [documentation site](https://github.com/tidyverse/dplyr/releases?page=1) and confirm that (1) Jan 2021 is missing from the plot, (2) all the version releases aren't getting plotted any more, and (3) the **Version Maturity** infoBox shouldn't be `NA`!

![image](https://user-images.githubusercontent.com/17878953/150548107-710a0a91-652b-41da-a64b-5f98bbea4e83.png)

I've traced the issue back to  `metric_cum_Info_upload_to_DB()` function which is responsible for generating & inserting community usage metrics into the database. To be specific, if you run this:

```
metric_cum_Info_upload_to_DB(package_name = "dplyr")
chk <- db_fun("SELECT * FROM CommunityUsageMetrics WHERE cum_id ='dplyr'")
```
And inspect `chk`, you get the below data: 

![image](https://user-images.githubusercontent.com/17878953/150546947-62dce26b-b00f-4711-a364-9b12e46b4b67.png)

First, I know why the "same month last year" row (in this case, that's Jan 2021, since the query was run Jan 2022) is missing. I apologize, but I think I missed it when I was revamping the code to go back 2 years in time instead of one year. Basically, we just need to remove/ comment out [this line](https://github.com/pharmaR/risk_assessment/blob/43bb66fa4b8cc7b542b541c0aa9280c504e9a555/Modules/dbupload.R#L219) of code. 

Secondly, two columns above are completely `NA`! I've walked through the function's code line-by-line (a few times) inside the `tryCatch` and when I do so, the code seems to produce the correct result (as seen below) so I'm not entirely sure why it doesn't work when calling `metric_cum_Info_upload_to_DB("dplyr)`. @marlycormar or @Robert-Krajcik, could one or both of you check it out?

![image](https://user-images.githubusercontent.com/17878953/150552895-ba1e4dc0-3eeb-439d-9770-19a7b6fd12b8.png)

**Some extra notes:**
1. Ignore the **position** column values... I was playing with something. You may have different values and it doesn't really matter. We don't use this column for anything in the app or the plot.

[...]

PR #165, opened 2022-01-25
re-write function metric_cum_Info_upload_to_DB

Robert-Krajcik:
Addresses issue #163


Robert-Krajcik:
So this is what the CommunityUsageMetrics graph looks like for dplyr now:
image
The downloads from last year are a bit higher because the logic used to remove the current month actually removed all the months with that name (all the January months).

marlycormar:
Closing this PR as it is no longer relevant.

Issue #168, opened 2022-02-15
Fix final decision submission


marlycormar:
Submitting final decision fails.

PR #173, opened 2022-02-17
overall comments yes/edit/no working, final decision working

Robert-Krajcik:
overall comments yes/edit/no working, final decision working
Addresses issue #168

### DIFF
--- a/R/sidebar.R
+++ b/R/sidebar.R
@@ -293,6 +293,11 @@ sidebarServer <- function(id, uploaded_pkgs, user) {
           WHERE name = '{selected_pkg()$name}'")
       )
       
+      disable("decision")
+      disable("submit_decision")
+      disable("overall_comment")
+      disable("submit_overall_comment")
+      
       removeModal()
       
       loggit("INFO",
--- a/R/sidebar.R
+++ b/R/sidebar.R
@@ -55,7 +55,7 @@ sidebarUI <- function(id) {
   )
 }
 
-sidebarServer <- function(id, uploaded_pkgs) {
+sidebarServer <- function(id, uploaded_pkgs, user) {
   moduleServer(id, function(input, output, session) {
     
     # Required for shinyhelper to work.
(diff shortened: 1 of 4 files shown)


marlycormar:
Hi @Robert-Krajcik, thanks for working on this issue! 

I jus tested the PR and my only comment so far is that after submitting the final decision the corresponding ui should be disabled (idem the final comment).

Robert-Krajcik:
Ready for you to review this again, @marlycormar

R/sidebar.R
```diff
@@ -185,39 +185,50 @@ sidebarServer <- function(id, uploaded_pkgs) {
                No - Exits from window and removes the text in comment box."
           ),
           footer = tagList(
-            actionButton("submit_overall_comment_yes", "Yes", class = "btn-success"),
-            actionButton("submit_overall_comment_edit", "Edit", class = "btn-secondary"),
-            actionButton("submit_overall_comment_no", "No", class = "btn-unsuccess")
+            actionButton(NS(id, "submit_overall_comment_yes"), "Yes", class = "btn-success"),
+            actionButton(NS(id, "submit_overall_comment_edit"), "Edit", class = "btn-secondary"),
+            actionButton(NS(id, "submit_overall_comment_no"), "No", class = "btn-unsuccess")
```
AARON-CLARK, line 190: Oh looky here! I've never actually used `NS()` on each UI id, but I see that's what Hadley is doing in Mastering Shiny. Previously, we always started any UI script with: 

`ns <- NS(id)`

and then called `ns()` in it's place, so line 188 above would start out like...

`actionButton(ns("submit_overall_comment_yes"), "Yes", class = "btn-success"),`

Going forward, do we want to use little ns or big ns?

AARON-CLARK (approved):
Looks great Robert! I just asked one small question regarding using little or big ns, but I'm happy to merge either way. Thanks for the PR!

R/sidebar.R
Robert-Krajcik: As long as we're consistent, @AARON-CLARK 
Yes, I'm OK with using little ns.

Robert-Krajcik: Actually, when I do
image
I find all the other functions are using big NS()
So maybe we should just leave it be.

AARON-CLARK: Let's leave it as-is then!

Issue #167, opened 2022-02-15
Add back admin view


marlycormar:
Add back the admin view that allows to add/delete users (shiny manager). Create a dedicated tab for this.

Robert-Krajcik:
This seems to work:

```{r, eval=FALSE}
            tabPanel(
              id = "admin_mode_id",
              title = "Administrator mode",
              # only using a div seems to work here
              div(id = "admin_mode",
              shinymanager:::admin_ui(id="admin")  )
            )
```
and then
`    if (user$role == "user") shinyjs::hide("admin_mode") `
to hide the admin functions from users

PR #175, opened 2022-02-18
put admin_mode on a tabset

Robert-Krajcik:
addresses issue #167

### DIFF
--- a/app.R
+++ b/app.R
@@ -83,8 +83,15 @@ ui <- fluidPage(
               id = "reportPreview_tab_id",
               title = "Report Preview",
               uiOutput("report_preview")  # UI for Report Preview tab Panel
+            ),
+            tabPanel(
+              id = "admin_mode_id",
+              title = "Administrator mode",
+              # only using a div seems to work here
+              div(id = "admin_mode",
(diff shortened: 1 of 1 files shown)


Jeff-Thompson12:
@Robert-Krajcik Including `admin_ui()` in the main UI of the application impacts the security of the application. The `fab_button` created by `shinymanager::secure_server()` has an input value of `.shinymanager_admin`. This input has the following observe event attached to it:
```
observeEvent(session$input$.shinymanager_app, {
      token <- getToken(session = session)
      updateQueryString(queryString = sprintf("?token=\"%s\"&language=\"%s\"", 
          token, lan()$get_language()), session = session, 
          mode = "replace")
      .tok$reset_count(token)
      session$reload()
  }, ignoreInit = TRUE)
```
This reloads the entire application. If a user tries to manually adjust the tokens to access the Admin panel, they will instead be redirected to the login screen. By just hiding the tab, a savvy user could reveal that tab even without administrative privileges and bypass the security logic of the `shinymanager` package. In fact, I was able to do just that with a non-admin user.

Robert-Krajcik:
Replaced by PR #204

Issue #177, opened 2022-02-22
uploadPackage doesn't show correct package version


Robert-Krajcik:
Currently the summary of uploaded packages shows the version as read from the csv file
![image](https://user-images.githubusercontent.com/75215958/155138700-c8135d1f-c1ed-4e02-ae21-ce19fa6072b0.png)
instead of the correct version (which is shown in the sidebar).
It should have shown this:
![image](https://user-images.githubusercontent.com/75215958/155138859-7acd05ab-7544-49b9-951d-b64147b70d2e.png)

PR #178, opened 2022-02-22
upload Package now shows correct version

Robert-Krajcik:
addresses issue #177


Robert-Krajcik:
This PR can do double-duty and address issue #49 as well.

Robert-Krajcik:
Suppose we have a .csv file like this:
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:///C:/Users/rkrajcik/AppData/Local/Temp/1/msohtmlclip1/01/clip.htm">
<link rel=File-List
href="file:///C:/Users/rkrajcik/AppData/Local/Temp/1/msohtmlclip1/01/clip_filelist.xml">
<style>

</style>
</head>

<body link="#0563C1" vlink="#954F72">



package | version
-- | --
gt | 0.0.0
LaOveja | 0.0.0
rvest | 0.0.0
ElCerdo | 0.0.0



</body>

</html>

When we read it in and process it with `uploadpackage.R,` we can check the source attribute
on its way out of `riskmetric::pkg_ref()` and issue a message if it is equal to `pkg_missing`

NOTE: package LaOveja was flagged by riskmetric as: 'pkg_missing' and will be removed. Did you misspell it?
NOTE: package ElCerdo was flagged by riskmetric as: 'pkg_missing' and will be removed. Did you misspell it?

Here is the summary page
image

Robert-Krajcik:
OK @marlycormar this one is ready for review!

Robert-Krajcik:
With regards to "internal" packages like stanly, the Summary of Uploaded Packages show stanly as "pkg_install"
image
However, there are some warnings:
```{r, eval=FALSE}
Warning in for (n in impnames) if (!is.null(genImp <- impenv[[n]])) { :
  closing unused connection 4 (https://cran.r-project.org/web/packages/stanly)
Warning in error_handler(x, ...) :
  no available scoring algorithm for metric of class "pkg_metric_error", returning default score of 0.
Warning in for (j in seq_along(x)) { :
  closing unused connection 4 (https://cran.r-project.org/src/contrib/Archive/stanly)
```
and then you get an error if you select the package.
```{r, eval=FALSE}
No rows were returned from query
SELECT * 
FROM community_usage_metrics
WHERE id = 'stanly'
Warning in min(year) : no non-missing arguments to min; returning Inf
Warning in min(month) : no non-missing arguments to min; returning Inf
Warning: Error in if: argument is of length zero
  1: runApp
```
So packages like stanly cannot be evaluated at this time.

Robert-Krajcik:
With a function like this, we could determine if the package was archived or not
```{r, eval=FALSE}
> get_archived <- function(cran = getOption("repos")){
+   if (is.null(cran)) cran <- "http://cran.rstudio.com/"
+   con <- gzcon(url(sprintf("%s/%s", cran, "src/contrib/Meta/archive.rds"), open = "rb"))
+   on.exit(close(con))
+   x <- readRDS(con)
+   names(x)
+ }
> pkg_name
[1] "stanly"
> pkg_name %in% get_archived()
[1] FALSE
> pkg_name <- "rvest"
> pkg_name %in% get_archived()
[1] TRUE
```

Robert-Krajcik:
Adding "stanly" to our .csv file, 
and with some additional coding, we get
```{r, eval=FALSE}
NOTE: package LaOveja was flagged by riskmetric as: 'pkg_missing' and will be removed. Did you misspell it?
NOTE: package ElCerdo was flagged by riskmetric as: 'pkg_missing' and will be removed. Did you misspell it?
NOTE: package stanly is not in the CRAN archive and will be removed.
```
and the Summary of Uploaded Packages contains known, archived packages only.
image

Server/uploadpackage.R
```diff
@@ -27,7 +27,40 @@ uploaded_pkgs <- reactive({
   
   names(uploaded_pkgs) <- tolower(names(uploaded_pkgs))
   uploaded_pkgs$package <- trimws(uploaded_pkgs$package)
-  uploaded_pkgs$version <- trimws(uploaded_pkgs$version)
+  
+  # https://stackoverflow.com/questions/33589591/check-if-package-name-belongs-to-a-cran-archived-package
+  get_archived <- function(cran = getOption("repos")){
+    if (is.null(cran)) cran <- "http://cran.rstudio.com/"
+    con <- gzcon(url(sprintf("%s/%s", cran, "src/contrib/Meta/archive.rds"), open = "rb"))
```
AARON-CLARK, line 34: I keep getting this error:

image


I think it's because my RStudio profile is pointing to more than one repo besides CRAN:

image

If you want CRAN only, you should request it explicitly by using: 
`cran = getOption("repos")[names(getOption("repos"))== "CRAN"]`

AARON-CLARK (changes requested):
Hi @Robert-Krajcik, this PR solves the issues mentioned after a few small fixes for me (detailed below). However, I think I found a useful use case that breaks your logic (unfortunately). I tried to upload a package called `samplesizeCMH` which is offers version 0.0.0 on CRAN. When I add it to my upload CSV, it does not upload to the app!

image

Another thing I think we should change right now: since we are not using the user provided versions in the CSV file for anything (correct?), can we please drop that requirement from the CSV file? It doesn't seem to be a supported feature with `riskmetric` either, so I we should get rid of it completely as it only confuses users. When `riskmetric` supports scoring risk for a certain package, we can bring it back.

Server/uploadpackage.R
```diff
@@ -9,8 +9,8 @@ uploaded_pkgs <- reactive({
   if(is.null(input$uploaded_file$datapath))
     validate('Please upload a nonempty CSV file.')
   
-  uploaded_pkgs <- read_csv(input$uploaded_file$datapath)
-  template <- read_csv(file.path('Data', 'upload_format.csv'))
+  uploaded_pkgs <- read_csv(input$uploaded_file$datapath, show_col_types = FALSE)
+  template <- read_csv(file.path('Data', 'upload_format.csv'), show_col_types = FALSE)
```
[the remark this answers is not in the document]
Robert-Krajcik: Yes, I tried `samplesizeCMH` and got
`NOTE: package samplesizeCMH is not in the CRAN archive and will be removed.`
But it really is a CRAN package.  So I have to use a different list.
The point is to only allow CRAN packages.
`riskmetric()` doesn't no nothing about `Stanly`

Server/uploadpackage.R
Robert-Krajcik: Doing this differently, won't need `getOption("repos")` anymore

Server/uploadpackage.R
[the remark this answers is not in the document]
Robert-Krajcik: updating DESCRIPTION via `use_this()`

Server/uploadpackage.R
Robert-Krajcik: I have updated this.
Now using this .csv
image

NOTE: package LaOveja was flagged by riskmetric as 'pkg_missing' and will be removed. Did you misspell it? 
NOTE: package stanly was not found in the CRAN archive and will be removed.

image

Robert-Krajcik: However, when you select on this particular package, you get
```{r, eval=FALSE}
No rows were returned from query
SELECT * 
FROM community_usage_metrics
WHERE id = 'samplesizeCMH'
Warning in min(year) : no non-missing arguments to min; returning Inf
Warning in min(month) : no non-missing arguments to min; returning Inf
Warning: Error in if: argument is of length zero
  1: runApp
Warning in min(downloads) :
  no non-missing arguments to min; returning Inf
Warning in max(downloads) :
  no non-missing arguments to max; returning -Inf
```

Robert-Krajcik:
Update one more time, this time using the same url that `dbupload.R` does to gather community usage metrics.
Now using the same .csv as above the messages are
```{r, eval=FALSE}
NOTE: package LaOveja was flagged by riskmetric as 'pkg_missing' and will be removed. Did you misspell it?                                                
NOTE: package stanly was not found in the CRAN archive and will be removed.
NOTE: package samplesizeCMH was not found in the CRAN archive and will be removed.
```
and the package summary:
image

AARON-CLARK:
Hi @Robert-Krajcik,

I saw you requested my review yesterday. Looking at your last comment & image posted on github, we won't be able to view `samplesizeCMH` in `riskassessment`? I think there may be other packages like it, so it would strengthen `riskassessment` overall if we could make it work... Did you change the URL because the the community usage error messages you received?

Robert-Krajcik:
I want this PR to work, @AARON-CLARK 
So I used the same URL as what community usage metrics is using.
What you are asking is just way beyond the scope of this PR.
If you want to review `samplesizeCMH` in `riskassessment` create another PR to address it.

AARON-CLARK:
> I want this PR to work, @AARON-CLARK So I used the same URL as what community usage metrics is using. What you are asking is just way beyond the scope of this PR. If you want to review `samplesizeCMH` in `riskassessment` create another PR to address it.

@Robert-Krajcik, I want this PR to work too! But I'm a little tentative to merge it since `samplesizeCMH` displays just fine in the `dev` branch. Since we don't know how many packages this PR could be excluding (beyond `samkplesizeCMH`) I think we need more info before choosing the best URL source. Maybe we can figure out why some pkgs exist on one list, but not the other? To figure out the scope of the issue, a good first step would be to find the number of packages available at the new vs old URLs, so we know how many packages are getting excluded.

AARON-CLARK:
So, I did a little research and it looks like there are 19,020 [available packages](https://cran.r-project.org/web/packages/available_packages_by_name.html) on CRAN and 18,229 [archived packages](https://cran.r-project.org/src/contrib/Archive/). So, this PR is potentially excluding ~791 legitimate R packages from being assessed in RAA because it leverages the archived URL.

I've confirmed that the package Archive  only features previous versions of the package or formerly available packages. So, it's basically missing all the packages that have only had one version release. As such, it doesn't make sense for RAA to rely solely on this list.

image

I'd like to get the open PR's pushed through, so if you don't mind @Robert-Krajcik, I can help work on this branch for a while. Was you're only concern with using the available pkg url these error messages? 

image

AARON-CLARK:
Also, I'm not sure if it matters, but the URL you were using for available packages was from [Rstudio](https://cran.rstudio.com/web/packages/available_packages_by_name.html)... it'd probably be better to use CRAN's url for consistency:

https://cran.r-project.org/web/packages/available_packages_by_name.html

PR #184, opened 2022-02-24
Fix package upload

marlycormar:
Uploading package failed due to incorrect variable name.

_Note:_ I took advantage of the PR to rename `package_name` -> `pkg_name` in the function `insert_maintenance_metrics_to_db`, although this instance was not causing any issues.

Fixes issue #184.


Robert-Krajcik (approved):
OK Looks good to me.

PR #186, opened 2022-02-28
add using SQLite vignette

Robert-Krajcik:


AARON-CLARK (changes requested):
Hi @Robert-Krajcik, Just a couple of requests:

First, I would recommend adding a small intro paragraph that addresses why your audience/ users should care about using SQLite at all. What does it do for them? How does it provide me value? Do they need to use it or is it optional? If it's required knowledge, why?

Then, I would suggest creating more vertical spacing between your paragraphs and before and after images. For example, I would throw 1 or 2 `<br>`s in all these locations:

image

AARON-CLARK (approved):
LGTM! Thanks @Robert-Krajcik

PR #193, opened 2022-03-02
sidebar minor update

Robert-Krajcik:
remove validate() from update package version step.

One thing I noticed was that 
when you select a package and then go and select "-"
the version does not get reset to "-"

This is because of this piece of code in the "update package version" step
```{r, eval=FALSE}
      if(input$select_pkg == "-")
        validate("Please select a package")
```

### DIFF
--- a/R/sidebar.R
+++ b/R/sidebar.R
@@ -77,7 +77,6 @@ sidebarServer <- function(id, uploaded_pkgs, user) {
       
       updateSelectizeInput(
         inputId = "select_pkg",
-        #label = h5("Select Package"),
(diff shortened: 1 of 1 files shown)


marlycormar:
Fixes the issue!

marlycormar:
Looking further into it, I notice that now it doesn't display the warning message on the metrics' tab when no package is selected.

Robert-Krajcik:
OK I can update this, @marlycormar

Robert-Krajcik:
Hi @marlycormar 
This has been updated now, and works like it did using fix_communityUsagePlot

marlycormar:
LGTM!

PR #195, opened 2022-03-03
Fix introjs guide

### LINKED ISSUE #172: Fix shiny intro JS
As the ui ids have changed, introJS doesn't work. We need to fix the introJS elements to match new app design.

AARON-CLARK:
This PR closes #172.

I had to create a couple of div()'s for certain UI's because it was too difficult to peg the UI element name or class. Otherwise, I tried to stick with the existing UI names where possible. I was also able to create a conditional set of steps for the uploadpackages UI, based on whether or not the user had uploaded a csv file (which reveals more UI objects).

A review from just one person will do. Thank you!

### DIFF
--- a/R/introJS.R
+++ b/R/introJS.R
@@ -26,8 +26,7 @@ introJSServer <- function(id, text) {
                   text %>%
                   union(sidebar_steps),
                 "nextLabel" = "Next",
-                "prevLabel" = "Previous",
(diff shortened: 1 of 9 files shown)


Robert-Krajcik (approved):
It's working now.  Thanks @AARON-CLARK 
image

Issue #198, opened 2022-03-05
 Cards have scroll bars on them which looks bad


marlycormar:

aclark02-arcus:
Super stale issue. Closing!

PR #199, opened 2022-03-05
Fix card scrolls & size - text is hidden

marlycormar:
Removed card scrolls. 

_Note:_ The cards (by bootstrap design) have a horizontal line that overflows from the cards. I think it doesn't look bad, but let me know if this is not the case to find a way to change it.


marlycormar:
Closes issue #198.

AARON-CLARK (changes requested):
Hi @marlycormar, thanks for the PR! The good news is the scroll bars are gone! But some other elements of the cards are getting pushed around. See this example when "dplyr" is selected:

image

AARON-CLARK:
@marlycormar, follow up to post above: I made sure my browser resolution was set to 100%. It must be because the URLs are wrapping to three lines instead of two. Can we truncate them slightly more? Or make the font waaaay smaller for URLs and emails? It is quite large right now.

marlycormar:
Thanks for reviewing this PR!

I'll cut the url string a bit more and make a few other changes to make it look better.

Robert-Krajcik:
Por supuesto, @marlycormar

Robert-Krajcik:
I'm using google chrome

AARON-CLARK:
@marlycormar, can you give this another look this week? I think you were really close the last time I reviewed the code:

> It must be because the URLs are wrapping to three lines instead of two. Can we truncate them slightly more? Or make the font waaaay smaller for URLs and emails? It is quite large right now.

AARON-CLARK:
closed with improved PR #296

PR #204, opened 2022-03-09
Admin mode fixes

Jeff-Thompson12:
So far I have used a similar approach as @Robert-Krajcik. Instead of using `shinyjs::hide`, I have used a render UI based on if the user was authorized as an administrator. This places the logic for displaying the admin UI in the server logic as opposed to the interface. This keeps users from being able to access the admin tab by editing the HTML or running JavaScript.

### DIFF
--- a/app.R
+++ b/app.R
@@ -121,10 +121,37 @@ add_tags <- function(ui, ...) {
   ui <- force(ui)
   
   function(request) {
+    query <- parseQueryString(request$QUERY_STRING)
+    admin <- query$admin
+    
   if (is.function(ui)) {
     ui <- ui(request)
   }
-  tagList(ui, tags$script(HTML("document.getElementById('admin-add_user').style.width = 'auto';")))
+  
+  if (identical(admin, "true")) {
+    tagList(ui, 
+            tags$script(HTML("document.getElementById('admin-add_user').style.width = 'auto';")),
+            tags$script(HTML("var paragraphs = Array.prototype.slice.call(document.getElementsByClassName('mfb-component--br'), 0);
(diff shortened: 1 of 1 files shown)


Jeff-Thompson12:
@Robert-Krajcik 
Just verifying that adding `options(shiny.fullstacktrace = TRUE)` is causing the PR to crash

Robert-Krajcik:
The admin tab itself looks great.
Except the other tabs won't open
image

Jeff-Thompson12:
@marlycormar 
With this PR I left the Administrative Mode link in the floating action button. Would you like me to remove it so that the Administrative Tools are accessible only in the tab?

The Administrative Mode in shinymanager also has a logs tab which shows number of logins and time spent in the app for different users. So it is not 1-1 with what is included in the Administrative Tools tab.

Robert-Krajcik:
Switching reviewer to @AARON-CLARK as I'm OOO this week

AARON-CLARK (approved):
@Jeff-Thompson12, I love it. I do have a few comments, but perhaps they make more sense as new issues. Let me know if you want to take care of them now or later:

When you update a user's info (like making them an admin for example), this green bar appears at the top and you literally can't click anything else while it's being displayed (which is about 4 or 5 seconds). Could we make it display for like 1 second max?
image

I clicked to edit multiple users at once, which is why these are the only fields visisble (below). Could we change the labels on these inputs to be "Start Date" and "Expire Date" or something? With the boxes being empty, its not obvious what these boxes are for until you click them to find the calendar pop-out. Like, expire could be TRUE / FALSE if the admin didn't know better (would probably be a bad admin if they didn't know... but still - let's make it really easy to understand when possible).
image

Issue #197, opened 2022-03-03
App crashes with select packages


AARON-CLARK:
If you upload `samplesizeCMH`, it will load into the app just fine, but if you select it from the pkg dropdown list, the app will crash. We are thinking this has to do with the the community usage tab because `samplesizeCMH` doesn't exist as an archived package. But if we turn on shiny's trace feature, I'm sure we'll find the bug:

`options(shiny.fullstacktrace = TRUE)`

AARON-CLARK:
closed with #206

PR #206, opened 2022-03-10
stop community usage from crashing

Robert-Krajcik:
addresses issue #197 App crashes with select packages

### DIFF
--- a/R/communityMetrics.R
+++ b/R/communityMetrics.R
@@ -8,10 +8,14 @@ communityMetricsServer <- function(id, selected_pkg, community_metrics, user) {
     # Render Output UI for Community Usage Metrics.
     output$communityMetrics_ui <- renderUI({
       
+      vect <- dbSelect("select distinct id from community_usage_metrics") %>% pull()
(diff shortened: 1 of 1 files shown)


AARON-CLARK (approved):
Looks like a good band-aid @Robert-Krajcik! This PR definitely stops the app from crashing, so I will merge it, but we'll still have to fix the root issue of #197, which is to have `insert_community_metrics_to_db()` use multiple URL sources to find pkgs before pulling community usage data. Right now, it's only using the CRAN archive to find packages, but as we know, that doesn't include ~800 packages that aren't in the archive because they only have 1 release.  I'll create a new issue and assign myself. The solution should leverage CRAN's list of 'available pkgs by name' in some way so that it finds a package like `samplesizeCMH`, and then grab downloads, etc. Here's proof `samplesizeCMH` has community usage: 100 downloads in the last week!

image

Robert-Krajcik:
Yeah this was a quick fix @AARON-CLARK 
I agree, we need to improve `insert_community_metrics_to_db()`

PR #218, opened 2022-03-21
Minor Package Upload revamping to improve Community Metrics

### LINKED ISSUE #215: Need to grab community usage metrics from more than 1 source
`samplesizeCMH` doesn't exist as an archived package on CRAN (only 1 release). There are ~791 other packages with only 1 release, but they still have lots of community usage. Right now, the app claims: 

image

However, that's not true. To prove it:

image

We'll need to add cran's '[available pkgs by name](https://cran.r-project.org/web/packages/samplesizeCMH/index.html)' into the solution. Notably, amending the `insert_community_metrics_to_db` function.

### LINKED ISSUE #217: `Latest Release` card is misspelled and incorrect
image

AARON-CLARK:
This PR touches 3 files, all to improve the package upload process from a Community Metrics perspective. Specifically:

* Fixes #215. Re-wrote some of `insert_community_metrics_to_db()` so that it could produce results for ~800 packages that have only a single version release (thus meaning, they don't exist on the CRAN Archive for webscraping)
* Added additional 'steps' to progress bar while packages were loading. Previously it was only 1 step per package, now there is like 5 incremental steps per package, giving the user a better idea of load time. I also improved the progress bar messaging slightly.
* Fixes #217

Just need one person to review and merge! Thank you!

### DIFF
--- a/R/communityMetrics.R
+++ b/R/communityMetrics.R
@@ -53,56 +53,43 @@ communityMetricsServer <- function(id, selected_pkg, community_metrics, user) {
       first_version <- community_metrics() %>%
         filter(year == min(year)) %>%
         filter(month == min(month)) %>%
-        slice_head(n = 1)
-
-      # Get difference between today and first release in years.
-      time_diff_first_version <- year(Sys.Date()) - first_version$year
-      first_version_label <- 'Years'
-      if(time_diff_first_version == 0){
-        # Get difference in months.
-        time_diff_first_version <- month(Sys.Date()) - first_version$month
-        first_version_label <- 'Months'
-      }
-      if(time_diff_first_version == 1)
-        first_version_label <- str_remove(
-          string = first_version_label, pattern = 's$')
+        slice_head(n = 1) %>%
+        mutate(fake_rel_date = lubridate::make_date(year, month, 15))
+      
+      # get the time span in months or years depending on how much time
+      # has elapsed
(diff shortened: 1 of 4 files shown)


Robert-Krajcik:
@Jeff-Thompson12 if this looks good to you, go ahead and approve it.   Thanks.

Jeff-Thompson12:
@AARON-CLARK couple of comments.

1. I tried including a random package (`abess` version 0.4.5) that had a new release today. The newest release is not included in community metrics.
2. Also, with the `abess` package, version 0.4.0 was released in December of 2021, but community metrics states the latest release was 1 year ago. (I don't think this observation is related to your PR though and maybe a separate issue should be created for it.)

AARON-CLARK:
Thanks @Jeff-Thompson12, I will look into both of these today. I wasn't a big fan of how the years were being reported either for # 2.

AARON-CLARK:
Okay @Jeff-Thompson12, following up to our convo:

1. So, I forgot, but the app never reports on any community metrics from the current month, since the month hasn't completed yet. I think the thought behind that design decision was because the number of downloads will always be lower for the current month, so it looks like the package is losing popularity. As such, you can't display a package's release from today.

2. I've improved how the time span is calculated using a new `utils.R` function called `get_date_span()` that I've applied to both cards on Community Metrics. Check it out!

image

Jeff-Thompson12 (approved):
@AARON-CLARK looks good to me!

PR #219, opened 2022-03-21
Allow admin to reset risk decision

Jeff-Thompson12:
Allows users with administrative roles to reset the risk decision for a selected package.

Additionally, I noticed that the reactive values returned from the sidebarServer module were only updating when a package is selected (i.e. they were not updated when a risk decision was submitted). I transitioned `selected_pkg` to be a `reactiveValues` element so that individual elements could be updated (e.g. `selected_pkg$decision` can be updated without reselecting the package).

### DIFF
--- a/R/sidebar.R
+++ b/R/sidebar.R
@@ -93,10 +93,7 @@ sidebarServer <- function(id, user, uploaded_pkgs) {
     # Get information about selected package.
     selected_pkg <- reactiveValues()
     
-    observe({
-      req(input$select_pkg)
-      req(input$select_ver)
-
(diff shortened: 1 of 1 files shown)


AARON-CLARK:
Hi @Jeff-Thompson12, there are some conflicts with the `dev` branch as it changed recently.

Jeff-Thompson12:
@AARON-CLARK 
Was expecting that but an easy fix.

AARON-CLARK:
That makes complete sense to me @Jeff-Thompson12. Thanks! Merging...

Issue #222, opened 2022-03-22
Admin tab improvements


AARON-CLARK:
If doable, some considerations for the Admin tab

1. When you update a user's info (like making them an admin for example), this green bar appears at the top and you literally can't click anything else while it's being displayed (which is about 4 or 5 seconds). Could we make it display for like 1 second max?
![image](https://user-images.githubusercontent.com/17878953/159512757-95f7b7eb-c52a-49a1-a6fe-3df80a7f5b95.png)

2. I clicked to edit multiple users at once, which is why these are the only fields visisble (below). Could we change the labels on these inputs to be "Start Date" and "Expire Date" or something? With the boxes being empty, its not obvious what these boxes are for until you click them to find the calendar pop-out. Like, expire could be TRUE / FALSE if the admin didn't know better (would probably be a bad admin if they didn't know... but still - let's make it really easy to understand when possible).
![image](https://user-images.githubusercontent.com/17878953/159514022-7afcff75-fbc3-48db-a910-f49ed43d245b.png)

PR #226, opened 2022-03-23
Improve administrative tools

Jeff-Thompson12:
This PR addresses #222.

It changes the label text for the modals created by `shinymanager` for creating and editing users.

Additionally it closes the notifications created by `shinymanager` after 1 second (or 1 second after the observe event is triggered). Unfortunately, since these notifications are not named, the solution was to use JS to return a list of all open notifications. At the current moment this does not seem to be an issue,

### DIFF
--- a/app.R
+++ b/app.R
@@ -127,7 +127,8 @@ add_tags <- function(ui, ...) {
   }
   
   if (identical(admin, "true")) {
-    tagList(ui, 
+    tagList(useShinyjs(),
+            ui, 
             tags$script(HTML("document.getElementById('admin-add_user').style.width = 'auto';")),
             tags$script(HTML("var paragraphs = Array.prototype.slice.call(document.getElementsByClassName('mfb-component--br'), 0);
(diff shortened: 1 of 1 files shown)


AARON-CLARK (commented):
HI @Jeff-Thompson12, I know you didn't ask me to review this yet, but can we change these column headers to be "Start Date" and "Expiration Date" too for consistency?

image

Also, if it's not asking too much, can this "Date" field have a better label too? Perhaps "Changed Date"?

image

Jeff-Thompson12:
@AARON-CLARK 
The last commit adds an observer for when those tables load and then fires JS to update the column names. I wanted to edit the column names before the data table displayed but couldn't find a way to make it work. Currently there is a split second before the column names update.

AARON-CLARK:
@Jeff-Thompson12 Excellent. That works for me. Let me know when you're done and I'll take a final look.

AARON-CLARK (approved):
Approved!

Issue #176, opened 2022-02-20
Potential features


marlycormar:
## Potential App Features
App Walk through:

#### General
- [x]  Make it supers obvious on the readme how to run the app for the first time (wait till app is a package) @marlycormar 
- [x] #198 (PR #199) - @marlycormar.
- [x] Selecting packages like "stanly" and "samplesizeCMH" cause the app to blow up.
We can either dis-allow these packages that don't have CRAN history, or fix the error and say
"Community Usage Metrics not available for this package" @Robert-Krajcik 
- [x] Click on MaintMetrics, select package, de-select package results in
`error: resulted in Error: Expecting a single string value: [type=character; extent=0]. @Jeff-Thompson12 
- [x] Report Preview into .docx format needs work @marlycormar @Robert-Krajcik 
- [x] Database overview -- download report fails (server problem) @AARON-CLARK 
- [x] Unable to see "admins_view" even though my role is admin  @Robert-Krajcik @marlycormar 
Not sure this is correct: req(input$res_auth$admin == TRUE)  -- This value is NULL 
- [x] The "CURRENT RISK SCORE WEIGHTS BY METRIC" is blank @Robert-Krajcik @marlycormar 

#### Uploaded packages 
- [x] When uploading a CSV, waitress loading bar for packages is not very informative about what package is getting accessed and if you upload a lot of packages your just left looking at percentages for a long time. Switch back to  @Robert-Krajcik 
- [x] #200 (PR #201) @marlycormar.
- [x] Table: Source field isn't self explanatory… data sourced from CRAN or Local? **Removing `source` for now** (PR #201)
- [x] Introjs not working if you upload a csv? @AARON-CLARK 

#### Sidebar
- [x] If you select a package and then select '-', you'll get an error: Expecting a single string value: [type=character; extent=0] (PR #203) @Jeff-Thompson12 
- [x] No visual feedback when you submit an initial overall comment and it doesn't update on the report preview tab when you submit a second overall comment  @Jeff-Thompson12 (PR #209)

#### Report preview
- [x] Download reports doesn't work there @AARON-CLARK 
- [x] Community usage metrics is missing @marlycormar 

#### Community Usage 
- [x] Make sure title and buttons don't overlap on the plot (fixed with #180).
- [x] Have a green color to display vertical lines?
- [x] Match the font? @marlycormar to send @AARON-CLARK the font name. (fixed with #180)

#### Database Overview
[...]

Issue #170, opened 2022-02-15
Complete report preview


marlycormar:
Report preview is missing the community usage metrics information. Address #169 and use the metric grid to display the community usage metrics info in the report preview.

aclark02-arcus:
Super stale issue. Closing!

PR #228, opened 2022-03-25
Downloads

AARON-CLARK:
Both download buttons for Report Preview AND Database overview should work for both HTML and docx. Please test both! I took some liberties consolidating some code if it was used more than 1 time in the app.

Partially related to #170 and written all over #176.


Jeff-Thompson12 (changes requested):
Pretty, pretty good. I like the functional changes. Both HTML and DOCX downloads are working on Report Preview and Database Overview. Couple of comments/change requests:

1. If only one package is selected on the Database Overview, can the download handler only output the one HTML or DOCX instead of a zip folder?
2. There is an x-axis issue with the number of downloads plot for the DOCX output. 
image
3. I kind of like the table output for comments in the HTML output over the DOCX. Could the DOCX at least bold the metadata components of the comment?
image
4. This change should probably be a separate PR, but I hadn't noticed until now the comments being created by the metric reweighting. I think I would prefer to have those comments as an additional overall comment instead of being added to both the Maintenance Metrics and Community Metrics.

AARON-CLARK:
@Jeff-Thompson12 Haha, can all these requests be a separate PR? I was just trying to get the code working again in this PR. Can you create new issues and assign me where necessary?

I love the suggestion for # 1. I think there is already an open issue for # 2. For # 4, I think the comments should exists for all comment types (Overall, MM, and Comm). However, we only display a single overall comment in the app's sidebar. So if there are more than 1 overall comments in the db, how do you display them all? Sounds like a slight UI design change may be necessary? Maybe we should ask the rest of the team.

Jeff-Thompson12:
@AARON-CLARK I'm fine with making them all a separate PR(s) since they are all cosmetic and this PR is fixing something broken in the application.

PR #237, opened 2022-04-01
Improve docx report

Jeff-Thompson12:
Some improvements to ReportDocx.Rmd.


AARON-CLARK:
@Jeff-Thompson12, is this ready for review?

Jeff-Thompson12:
@AARON-CLARK Yes

Jeff-Thompson12:
I changed where the Rmd files were being ran because it made it "easier" to include the icons in the docx, but if the Rmd files need to be ran in a temporary directory I can modify the code to handle that.

AARON-CLARK (changes requested):
This appears to work for me (locally). However, copying files to a temp dir is important when the app is deployed on a server (like shinyapps.io), so I think that code needs to be brought back. Here is a snippet from Wickham's "Mastering Shiny" book:

image

AARON-CLARK (approved):
LG

PR #251, opened 2022-07-15
Further Packagization

AARON-CLARK:
A work in progress... but this PR attempts to fully golemize the `dev` branch. The original reason for doing this is (1) using `{golem}` is a best practice and (2) we wanted to fully packagize the app so that we could meet other goals, like...

* using `{pkgdown}` to create a better documentation site
* make it easier for users to download/install the app for users via `remotes::install_github("pharmaR/risk_assessment")`
* be able to export useful visuals like our metric cards and community usage plot
* make a hex, because we love hex logos

### Todo
- [x] Add typical `{golem}` files and adhere to their file structure
- [x] Add `{roxygen}` code to all modules
- [x] Add `{roxygen}` code to all functions
- [x] Call out pkg name prefix in all function calls (ex. `dplyr::mutate()`)
- [x] Make sure reports work  --these should be good now JT
- [x] community usage metrics are no longer flowing through  --I'm not having any issue JT
- [x] Make `rintrojs::introjs()` work again (low priority)  --Also not noticing where `rintrojs::introjs()` is not working JT
- [x] Make it work by running `dev/run_dev.R`  --this is up and running now JT

### Note
When running `dev/run_dev.R`, I get the following error - which is clearly the most well documented vague and useless error message R can produce:

image

I'm guessing it has something to do with how `{golem}` and `{shinymanager}` don't play well together, but it could be something else. So, in order to make progress in development, I've been running everything `dev/run_dev.R` except `run_app()`. Instead, I modified the `R/global.R` file to not `library()` all packages so that it has to rely on the `golem::document_and_reload()` to find the necessary imports. THEN, I just run `app.R`. Then I just play around in the app till something breaks due to an undeclared import. But lately, I've found it challenging to find any errors, with the exception of the to-do list above.

Unfortunately, this PR has become a bottleneck holding up other development of RAA, and there is a lot of new things coming down the pipe.


AARON-CLARK:
This may be why the reports are rendering anymore, but it appears that community usage metrics are no longer flowing through the that tab!

image

Jeff-Thompson12:
@AARON-CLARK I have not gone through the changes line by line yet, but I pushed a few commits that fixed a few of the outstanding issues with the packagization (such an ugly word).

Jeff-Thompson12 (commented):
@AARON-CLARK I've looked through the changes and played around in the application and everything looks good to me. Of course I also pushed my own commits, so let me know what you think about the changes.

AARON-CLARK (commented):
Hi @Jeff-Thompson12, thanks for all the work on this. Right now, I noticed these things right away on your last 10 commits:

1. no Rd file was generated for this function. I pushed a commit ([929dda7](https://github.com/pharmaR/risk_assessment/pull/251/commits/929dda7edf29677d99ddf2d6f54285619e1c3159)) to add it.

image

2. I can't test the app due to this error received upon running `run_dev.R`: 

image

AARON-CLARK:
@Jeff-Thompson12, sorry I sent that without doing any research. It looks like running `bslib::bs_theme()` creates file attachments locally, so I don't think we'll be able to store `app_theme` as internal data:

image

image

I can push a commit to fix it since I want to test the app anyhow. Thanks!

Jeff-Thompson12:
@AARON-CLARK I see. Doesn't sound like an easy fix. Even putting it in an R file would impact being able to put the application in a package. Interested in seeing the fix you push.

AARON-CLARK:
> Doesn't sound like an easy fix. Even putting it in an R file would impact being able to put the application in a package. Interested in seeing the fix you push.

@Jeff-Thompson12, I'm not sure what you mean. I put it an an R file (utils_initialize.R) and it seems to be working. However, perhaps you could test it? Only trouble is we lost the blue background on the login screen, but it seems like we retained all the other bootstrap 5 theming... I think?

image

AARON-CLARK:
## Todos:
- [ ] Where did the blue background go on the login screen?
- [ ] Does the app install correctly on Jeff's machine (with correct paths to "attached files" generated from the `app_theme` object - aka `bslib::bs_theme()`  given it was built on Aaron's machine?

Jeff-Thompson12:
@AARON-CLARK & @Robert-Krajcik 

I was fiddling and randomly the blue background returned for me. I do not know why it went away or why it returned. The formats are located in the CSS file login_screen.css. I would be interested in if @Robert-Krajcik has a blue login screen or white.

AARON-CLARK:
@Jeff-Thompson12, still white for me! @Robert-Krajcik, what do you see?

image

Jeff-Thompson12:
@AARON-CLARK & @Robert-Krajcik,

I just pushed a commit that should fix the login screen. My understanding is that when running the application with app.R we were bypassing some settings in `{shinymanager}`. If you inspect the document, you can see that the file is not accessible in the login screen but becomes available once the user is logged in. I have gotten around this issue by adding the style in as text instead of linking to a file.

AARON-CLARK (commented):
Blue is back! Thanks @Jeff-Thompson12. I think this PR is ready for a final review.

Robert-Krajcik (changes requested):
I gave this app a good workout, just noticed this one thing:
I make a decision about a package yet the database tab doesn't reflect that change.
It still thinks no decision was made
image
image

Jeff-Thompson12:
Nice catch. This doesn't appear to be a new issue either. The deployed version on shinyapp.io has the same problem.

Jeff-Thompson12:
While working on fixing this I also came across a bug where the slider becomes "reenabled" when you select a different package despite a decision being made.

image

Jeff-Thompson12:
@AARON-CLARK & @Robert-Krajcik 

I just pushed two commits to address both issues.

AARON-CLARK (commented):
Looks great. Thanks @Jeff-Thompson12

PR #274, opened 2022-11-03
Next Release

AARON-CLARK:


Jeff-Thompson12:
@AARON-CLARK These checks appear to be failing because `riskmetric` could not populate the downloads field. I think this raises an interesting question for our application. What happens if something fails when adding a package to the assessment database? We currently do not have any convenient way to delete and re-add or reload a package.

Issue #257, opened 2022-08-30
Refresh readme: how to run the app for the first time


AARON-CLARK:
- [x] Be very descriptive & update the wording on the readme
- [x] The app may be case sensitive when it comes to username credentials. For now, just add alternative user names that are all lowercase and all uppercase to solve most people's issues.
- [x] Allow user executing `run_app()` the ability to change both login notes and package version.

See PR # for changes:

![image](https://user-images.githubusercontent.com/17878953/201964995-94a1a044-5a38-4390-a1fd-a323e99fb23e.png)

![image](https://user-images.githubusercontent.com/17878953/201965074-b92d35a7-de82-498c-b318-275f7a666947.png)

But by default, with `run_app()`:

![image](https://user-images.githubusercontent.com/17878953/201965774-b1536b39-5ace-4d55-b730-5234d3aba461.png)

AARON-CLARK:
I think we should have a vignette dedicated to managing the credential database... so I initially wrote this out in the README, but decided to save it for later.

Once an admin account has been created, we advise the admin(s) manually create accounts for new users in the app. You can disable the self-service approach to role assignment using `allow_auto_roles = FALSE` in `run_app()`.

```{r, eval=FALSE}
riskassessment::run_app(allow_auto_roles = FALSE)
```

AARON-CLARK:
We need a vignette dedicated to deployments as well, that will probably expand upon the readme in this way:

AARON-CLARK:
closed with #279

Issue #278, opened 2022-11-15
New `run_app()` args


AARON-CLARK:
Including these in the same issue since they are somewhat related:
- [ ]  feature to run the app without the authentication wall, with the user role being 'admin' by default. This would just be used for development purposes as it would save us load of time while testing features. I envision there would be a new argument in run_app() called `pre_auth_user` that is `NULL` by default (meaning, you would have no pre-authorized user... you'd have to log in like normal), but when "admin" or "nonadmin" is specified, it would create or use some a username called 'default_admin' or 'default_nonadmin' in the credentials database that wouldn't ask the user to change the password and would ideally bypass {shinymanager} auth screen altogether.

- [ ] Allow users to disable the self service approach to role assignment (the process using QWERTY1 password) using allow_auto_roles = FALSE in run_app().

AARON-CLARK:
old / stale, but complete!

PR #279, opened 2022-11-15
Improve doc and experience for first time running the app

AARON-CLARK:
See #257

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -124,11 +124,11 @@ create_credentials_db <- function(db_name = golem::get_golem_options('credential
   
   # Init the credentials table for credentials database
   credentials <- data.frame(
-    user = c("ADMIN",  "NONADMIN"),
-    password = rep("QWERTY1", 2),
+    user = "ADMIN",
+    password = "QWERTY1",
     # password will automatically be hashed
-    admin = c(TRUE, FALSE),
-    expire = rep(as.character(Sys.Date()), 2),
+    admin = TRUE,
+    expire = as.character(Sys.Date()),
     stringsAsFactors = FALSE
   )
   


README.Rmd
```diff
@@ -13,77 +13,118 @@ knitr::opts_chunk$set(
...
-The Risk Assessment App is an interactive web application serving as a front end
-application for the [`riskmetric`](https://github.com/pharmaR/riskmetric) R package. `riskmetric` is a framework to quantify risk by assessing a number of
-metrics meant to evaluate development best practices, code documentation,
-community engagement, and development sustainability. The app and `riskmetric`
-aim to provide some context for validation within regulated industries.
+`riskassessment` is a shiny app providing a handy interface which augments the utility of the [`riskmetric`](https://github.com/pharmaR/riskmetric) package within an organizational setting.
 
-Furthermore, the app extends the functionalities of `riskmetric` by allowing
+<center>
+<a href="rinpharma.shinyapps.io/risk_assessment">
```
Jeff-Thompson12, line 25: Link doesn't work. Perhaps just missing `https://`?

README.Rmd
```diff
@@ -13,77 +13,118 @@ knitr::opts_chunk$set(
...
+riskassessment::run_app()
+```
 
-We would like to thank all the contributors!
 
-Specially, we would like to thank the following contributors/authors.
+
+#### Deployments
+
+As you might expect, certain deployment environments offer persistent storage and others do not. For example, shinyapps.io does not. That means that even our [demo app](rinpharma.shinyapps.io/risk_assessment) that's hosted on shinyapps.io contains a a package database that can't be permanently altered. That's not advantageous since an organization needs to continually add new packages, publish comments, and make decisions about packages. Thus, we'd recommend exploring these deployment options (which allow persistent storage):
```
Jeff-Thompson12, line 94: Same note as above. Link doesn't work.

R/utils.R
```diff
@@ -124,11 +124,11 @@ create_credentials_db <- function(db_name = credentials_name){
...
   credentials <- data.frame(
-    user = "admin",
-    password = "QWERTY1",
+    user = c("ADMIN", "admin", "NONADMIN", "nonadmin"),
+    password = rep("QWERTY1", 4),
     # password will automatically be hashed
-    admin = TRUE,
-    expire = as.character(Sys.Date()),
+    admin = c(TRUE, TRUE, FALSE, FALSE),
+    expire = rep(as.character(Sys.Date()), 4),
```
Jeff-Thompson12, line 131: Personally I'm not a big fan of adding additional users to the initialized credential database. It is not clear to the user that multiple admin account were initialized and that they would need to open admin tools to remove the additional accounts.

image

R/run_app.R
```diff
@@ -1,29 +1,41 @@
 #' Run the Shiny Application
 #'
-#' @param ... arguments to pass to golem_opts.
-#' See `?golem::get_golem_options` for more details.
+#' @param app_ver a "global" variable that is passed to several modules &
```
Jeff-Thompson12, line 3: I think maybe we should be using `{golem}` [configurations](https://cran.r-project.org/web/packages/golem/vignettes/e_config.html) to handle these "global" variables. Then we can be utilizing functions like `get_golem_config()`. This seems like a better approach than using `assign()`.

R/run_app.R
```diff
@@ -1,29 +1,41 @@
...
+#' @param ... arguments to pass to golem_opts. See `?golem::get_golem_options`
+#'   for more details.
 #' @inheritParams shiny::shinyApp
 #'
 #' @export
 #' @importFrom shiny shinyApp
 #' @importFrom golem with_golem_options
 run_app <- function(
-  onStart = initialize_raa,
+  onStart = initialize_raa(),
```
Jeff-Thompson12, line 16: The parentheses `()` should not be present. This caused my app to error out on a fresh initialization.

R/run_app.R
```diff
@@ -1,29 +1,41 @@
...
   options = list(),
   enableBookmarking = NULL,
   uiPattern = "/",
+  app_ver = paste(packageVersion("riskassessment")),
+  login_note = HTML('<em>Note:</em> To log in for the first time, use either the admin user:
+                        <u>admin</u> or the non-admin user: <u>nonadmin</u>, both with
+                        password <u>QWERTY1</u>.'),
```
Jeff-Thompson12, line 23: I prefer handling this in the function body itself. It also allows the opportunity to only show the `login_note` if the credentials database has not been created yet.

[run_app.txt](https://github.com/pharmaR/risk_assessment/files/10015742/run_app.txt)

Jeff-Thompson12 (changes requested):
@AARON-CLARK I like the idea of the new changes, especially the updated README. I do think we should avoid using `assign()` though. We should be able to handle this issue with `{golem}` options, but unfortunately that would require some recoding elsewhere as well.

R/run_app.R
Jeff-Thompson12: Actually dug into this a little more, you can pass `app_version` as an option within `run_app()` for the argument `golem_opts` (i.e. `golem_opts = list(app_version = app_ver, ...)`). Then you can query this value in the code using `golem::get_golem_options('app_version')`.

A lot easier to implement than I thought. We should consider adding the database names and other values as `{golem}` options.

R/run_app.R
AARON-CLARK: Oh sorry. I actually was testing out adding a argument to this function called "ver" to handle the app_version stuff. It actually worked well, but I felt it was weird calling `run_app()` like this:

```
run_app(onStart = initialize_raa(ver = "my version"))
```
so I abandoned it but forgot to remove the `( )`.

R/utils.R
AARON-CLARK: That's a good point. Maybe we should just switch back to a single 'admin' and 'nonadmin' account for now. Ideally, to solve the case sensitivity issue, I'd like to be able to pass a `tolower` function to all usernames entered to avoid the confusion/frusteration. I'll investigate.

AARON-CLARK:
Hi @Jeff-Thompson12, I made some changes that are ready for your review. Love the `get_golem_options()` function. It's my first time using (or needing) that feature and it's really intuitive! Gosh, I love `golem`. The only thing that isn't resolved in my mind is the case sensitive usernames in `shinymanager`... Open to hear your thoughts. Or we can merge this AS-IS and punt that topic for another time.

Jeff-Thompson12:
> The only thing that isn't resolved in my mind is the case sensitive usernames in `shinymanager`... Open to hear your thoughts. Or we can merge this AS-IS and punt that topic for another time.

@AARON-CLARK I think we should punt on the case sensitive usernames for the time being. The problem is that an admin user can change those values using the administrative tools from `{shinymanager}` where we do not have control to pass a `tolower()`.

AARON-CLARK:
@Jeff-Thompson12, sounds good. Let me know when this looks good and I'll "bypass branch protections" to merge.

Jeff-Thompson12 (commented):
Only the one additional feedback.

R/run_app.R
```diff
@@ -12,24 +21,44 @@ run_app <- function(
...
+  credentials_db_name = NULL,
+  assessment_db_name = NULL,
   ...
 ) {
+  # Pre-process some run-app inputs
+  if(is.null(app_ver)) app_ver <- paste0(packageVersion("riskassessment"))
+  if(is.null(assessment_db_name)) assessment_db_name <- "database.sqlite"
+  if(is.null(credentials_db_name)) credentials_db_name <- "credentials.sqlite"
+  if(is.null(login_note) & !file.exists(credentials_db_name)) {
+    login_note <- HTML('<em>Note:</em> To log in for the first time, use the admin user:
```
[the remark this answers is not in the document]
AARON-CLARK: Yes, that actually was intentional. I was thinking that long term we'll probably only want to note the 'admin' account, since that's what you need to get started with the app... but I thought maybe we'd keep the 'nonadmin' account on there just for testing purposes? What do you think? Our future selves will probably be confused by this.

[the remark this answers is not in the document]
Jeff-Thompson12: Initial reaction is "a temporary solution becomes a long problem." It poses less risk than multiple admin accounts though. I'm good with merging as-is.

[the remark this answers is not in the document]
AARON-CLARK: eh, I'll just get rid of the nonadmin for now. Maybe when we tackle #278, we can create a 'default_admin' and 'default_nonadmin' account launched directly from `run_app()` for testing purposes (bypassing auth screen).

Jeff-Thompson12 (dismissed):

R/run_app.R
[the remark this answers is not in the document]
Jeff-Thompson12: I like that idea better. I think we have been under utilizing `{golem}` options and configurations. We should work to setup a development configuration.

AARON-CLARK:
@Jeff-Thompson12, Please provide (hopefully) one last review after my recent commit!

Jeff-Thompson12 (approved):
LGTM!

Issue #263, opened 2022-08-30
Clear out R CMD Check issues


AARON-CLARK:

Robert-Krajcik:
PR #282 addressed this issue.

PR #282, opened 2022-11-16
Fix rcmd checks

Robert-Krajcik:
Addresses Issue #263 Clear out R CMD Check issues


.Rbuildignore
```diff
@@ -8,3 +8,4 @@ $credentials.*
 $database.*
 ^README\.Rmd$
 ^\.github$
+^LICENSE\.md$
```
Jeff-Thompson12, line 11: It feels like we should add some items to `.Rbuildignore` to knock out a couple notes.

image

Jeff-Thompson12 (commented):
Is there any way to determine what functions (if any) are being used by `{crayon}`, `{desc}`, & `{forcats}`?

image

.Rbuildignore
Robert-Krajcik: Yeah I think I can find which functions from `crayon`, `desc` and `forcats` are being used here.
We're getting rid of `www`
And I can add some files to .gitignore

AARON-CLARK (dismissed):
Hi @Robert-Krajcik, looks great! Thanks for the PR. The only note I have is that all the changes you made the the `README.md` file will likely be overwritten the next time someone renders the `README.rmd` file. We may even start using a GitHub Action workflow to automatically render it for us. If you think you can improve the Rmd file, feel free to amend it in a separate PR. I'm going to merge this one!

PR #288, opened 2022-11-17
Fix Assessment Criteria tables not displaying

Jeff-Thompson12:
Don't pre-render `datatables` as internal data so the app is not dependent on local files.


AARON-CLARK (approved):
LGTM

PR #289, opened 2022-11-17
app.R initiated for deployment

AARON-CLARK:
This is in response to recent Q&A question: https://github.com/pharmaR/risk_assessment/discussions/287

Please review that use case!

I created an app.R file using 'golem::add_rstudioconnect_file()'. Which golem deployment function used isn't very important (there are three). Only downside is it did add `pkgload` as a new import... not sure how I feel about that.

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -8,3 +8,5 @@ $credentials.*
 $database.*
 ^README\.Rmd$
 ^\.github$
+^app\.R$
+^rsconnect$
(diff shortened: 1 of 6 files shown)


Jeff-Thompson12 (commented):
I am able to run the app using both `shiny::runGitHub('risk_assessment', 'pharmaR', ref='add_app_file')` and `shiny::runUrl("https://github.com/pharmaR/risk_assessment/archive/add_app_file.zip")` and it is showing expected behavior.

My only question is do we want this update to merge into `dev` or `master` since it is affecting current application usage?

AARON-CLARK:
> do we want this update to merge into `dev` or `master` since it is affecting current application usage?

Make sense to me!

AARON-CLARK:
No longer needed thanks to #294.

Issue #277, opened 2022-11-15
there are two `www/` folders in the app


AARON-CLARK:
The `www/` folder in the root dir should be deleted as it is outdated (used prior to golemizing). That is, we should use that found at `/inst/app/www/`. Need a PR to delete the root dir and test the app to make sure nothing else is relying on it.

PR #290, opened 2022-11-17
remove www folder from root directory

Robert-Krajcik:
Addresses issue #277 there are two www/ folders in the app

I have convinced myself by searching for all instances of www
that we are using `inst/app/www` and not `www`

And the app still runs!


Jeff-Thompson12 (dismissed):
LGTM!

AARON-CLARK (approved):
LGTM! I just pushed a quick commit that updates the dev version by 1 using `usethis::use_version("dev")`. We should get in the habit of doing that (just in case).

Issue #275, opened 2022-11-10
Report download not working with `{fontawesome}` Version `0.4.0`


Jeff-Thompson12:
The download reports are not working on MAC computers. This appears to be related to two issues.

1. The Community Metrics are not being created because of an empty select. (See #283)
2. The icons are causing problems when the Rmd file is being knit.

Problem 1 can be easily handled.
[comm_patch.txt](https://github.com/pharmaR/risk_assessment/files/9983788/comm_patch.txt)

Problem 2 might require a bit more pandoc knowledge.

AARON-CLARK:
Hey @Jeff-Thompson12, could you update this issue and separate it into two so that problem 1 could be solved with a PR? Thanks!

Jeff-Thompson12:
Problem one is being addressed in #283 

This issue is being left open until a solution to download HTML reports on Mac computers is resolved.

Jeff-Thompson12:
@AARON-CLARK this is not a Mac issue. It is a `{fontawesome}` package issue. Version `0.4.0` does not include files "fa-v4compatibility.tff" and "fa-v4compatibility.woff2" files. While you can manually search for these files online and add them to the appropriate folder, this is not desirable behavior.

Jeff-Thompson12:
See https://github.com/rstudio/fontawesome/issues/99

Jeff-Thompson12:
@imran3004 for the time being I would suggest loading an older version of `{fontawesome}`. Version `0.3.0` was working for me.

imran3004:
Yes `fontawesome` 0.3.0 is working on my Mac. I am able to generate the HTML reports.

AARON-CLARK:
will be solved with #48

AARON-CLARK:
Just thought of this: #48 will only resolve this issue for developers, not users. Thus, shouldn't we add `fontawesome (== 0.3.0)` to the suggest field until this issue is resolved with `{fontawesome}`?

Jeff-Thompson12:
@AARON-CLARK BTW, version 5.0 of `{fontawesome}` has been released and resolves this issue. Not sure if we want to leave the checks in place but maybe only provide the notes and modals if the user is using version 4.0. Thoughts?

AARON-CLARK:
@Jeff-Thompson12, that makes sense to me.

PR #292, opened 2022-11-17
Ih issue264

imran3004:
Verified that downloaded DOCX report and the preview of the report show the score. I was not able to verify this for the HTML report, because of issue #275. **Please check on a Windows machine.**


R/reportPreview.R
```diff
@@ -1,7 +1,7 @@
 #' UI for 'Report Preview' module
-#' 
+#'
```
Jeff-Thompson12, line 2: @imran3004 there are a lot of formatting edits in your PR which is making it hard to go through all the changes. Is this happening automatically on your end?

R/reportPreview.R
```diff
@@ -178,21 +178,23 @@ reportPreviewServer <- function(id, selected_pkg, maint_metrics, com_metrics,
...
+      req(selected_pkg$name())
+
       tagList(
         h5('Overall risk:'),
-        ifelse(selected_pkg$decision() == '', 
+        ifelse(selected_pkg$decision() == '',
                'Pending',
-               selected_pkg$decision()))
+               selected_pkg$decision()),
+        h5('Score:'), selected_pkg$score())
```
Jeff-Thompson12, line 191: In the Sidebar we have the score labeled as "Risk". I personally think the label should be "Risk Score:" over just "Score:" but am open to differing opinions.

Jeff-Thompson12 (changes requested):
@imran3004 looking good. Just a handful of things.

- [ ] The score is not being pulled in when downloading reports using the database tab.
image
- [ ] Can you merge the removed white space back into your branch? I think if we want to fix the white space formatting, it would be better to do that in it's own PR.
- [x] Make placement of the score consistent between the Report Preview and the report downloads.

AARON-CLARK:
Hi @imran3004, I just merged #282 which fixes all the R CMD Check warnings/errors. Looks like that may have created some merge conflicts with your branch as well. As such, I'd recommend pulling in the recent changes on `dev` and merging it locally with your `ih-issue264` branch and re-push! Thanks!

R/reportPreview.R
imran3004: @Jeff-Thompson12 I think that's because I am using Emacs as my editor. Let me use RStudio.

R/reportPreview.R
imran3004: @Jeff-Thompson12 I agree, the label Risk Score is better than just Score. I'll change this.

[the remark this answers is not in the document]
imran3004: @Jeff-Thompson12 yes, I like it placed above. As you read information from the top, the Risk Score is a number which is then converted to a grade, the Overall Risk.
I'll make the change.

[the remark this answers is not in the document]
imran3004: Changed. (Commit ID - a2226d7)

imran3004: Changed. (Commit ID - f0efc77)

R/reportPreview.R
imran3004: I have merged the `dev` branch into this branch and used RStudio to make changes. Can you please confirm if the formatting edits have been reverted?

R/reportPreview.R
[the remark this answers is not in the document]
imran3004: The Risk Score is now available in reports downloaded from the Database tab. (Commit ID - f1fa368)

imran3004:
Verified that HTML reports can be generated on macOS using `fontawesome 0.3.0` instead of the latest version of the package.

.Rbuildignore
```diff
@@ -8,9 +8,3 @@ $credentials.*
 $database.*
 ^README\.Rmd$
 ^\.github$
-^LICENSE\.md$
-^.open$
-^credentials.sqlite$
-^database.sqlite$
-^docs$
-^loggit.json$
```
Jeff-Thompson12, line 16: These lines should not be deleted from `.Rbuildignore`.

R/reportPreview.R
Jeff-Thompson12: @imran3004 the edits have not been reverted

Jeff-Thompson12: @AARON-CLARK do we care about white space edits? There is the option to hide them when reviewing file changes.

Jeff-Thompson12 (changes requested):
@imran3004 it appears that some lines adding to `.Rbuildignore` got removed in commit 8f373f3.  If Aaron is fine with the white space edits we can move on. I did find [this](https://gist.github.com/hello-josh/dbabb56e1a01ac293b83154a6f4a208e). I do not think it is as simple as merging `dev` back into the branch.

AARON-CLARK:
Hi @imran3004 & @Jeff-Thompson12, to make things less confusing for everyone, I'd prefer that only lines with actual code changes be included in PRs. There may be a time when formatting whitespaces is important for code readability, but not in this case. Hope that helps!

PR #293, opened 2022-11-17
Auto login

Jeff-Thompson12:
If given credentials, the app will automatically log the user in when running the application in development.

### DIFF
--- a/R/run_app.R
+++ b/R/run_app.R
@@ -42,6 +42,12 @@ run_app <- function(
       ui = add_tags(shinymanager::secure_app(app_ui,
         tags_top = tags$div(
             tags$head(tags$style(HTML(readLines(system.file("app", "www", "css", "login_screen.css", package = "riskassessment"))))),
+            tags$head(if (!get_golem_config("app_prod") && !is.null(golem::get_golem_options("login_creds"))) {tags$script(HTML(glue::glue("function pageLoaded() {{
+    Shiny.setInputValue('auth-user_id', '{golem::get_golem_options('login_creds')$user_id}');
+    Shiny.setInputValue('auth-user_pwd', '{golem::get_golem_options('login_creds')$user_pwd}');
+    document.getElementById('auth-go_auth').click();
+  }}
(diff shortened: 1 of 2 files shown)


AARON-CLARK (commented):
@Jeff-Thompson12, besides some merge conflicts on this branch, I can confirm this approach does indeed work. However, I think it'd be tedious to remember to remove our login credentials from `run_dev.R` before pushing the repo back to GitHub. I shared a slightly different vision for how this would execute in #278. Specifically:

> I envision there would be a new argument in run_app() called `pre_auth_user` that is `NULL` by default... when `pre_auth_user = 'admin'` (or 'nonadmin'), it would **create** a username called 'default_admin' (or 'default_nonadmin') in the credentials database...

... and login with username = 'default_admin' and password = '[insert generic pwd]' similar to this approach. And maybe on app exit, it would remove it from the database to keep things tidy? Thoughts, comments, criticisms?

Robert-Krajcik:
HI @AARON-CLARK and @Jeff-Thompson12 
Is there some way for the user to set an option
like `options(pre_auth_user = TRUE)` and not touch the run_app() at all?
Then any admin on the development app would have their user_id an password plugged in for them.

Jeff-Thompson12:
@AARON-CLARK I can envision a scenario where when running the app in development mode with `pre_auth_user = "admin"`, a separate credentials database is used. Maybe named `credentials_dev.sqlite` that is populated with an admin and non-admin user. I can work on creating something like that.

@Robert-Krajcik There are a number of ways that global options could be set for the application. What do you see as the benefit over passing them as `{golem}` options?

AARON-CLARK (changes requested):
@Jeff-Thompson12, this is amazing. A PR truly beyond my wildest dreams. Myself and all future developers of `riskassessment` are indebted to you!

Only one request - `dev` is currently v0.0.1.9008. Could you increment one higher please?

AARON-CLARK (approved):
LGTM

PR #294, opened 2022-11-21
app.R initiated for deployment for `master` branch

AARON-CLARK:
This is in response to recent Q&A question: https://github.com/pharmaR/risk_assessment/discussions/287

Please review that use case!

I created an app.R file using 'golem::add_rstudioconnect_file()'. Which golem deployment function used isn't very important (there are three). Only downside is it did add pkgload as a new import... not sure how I feel about that.

Related to #289

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -8,3 +8,5 @@ $credentials.*
 $database.*
 ^README\.Rmd$
 ^\.github$
+^app\.R$
+^rsconnect$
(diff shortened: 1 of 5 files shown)


AARON-CLARK:
How does this one look @Jeff-Thompson12?

Jeff-Thompson12 (approved):
LGTM!

I am able to run the app using both `shiny::runGitHub('risk_assessment', 'pharmaR', ref='app_file_master')` and `shiny::runUrl("https://github.com/pharmaR/risk_assessment/archive/app_file_master.zip")` and it is showing expected behavior.

PR #296, opened 2022-11-22
Fix card scrolls

### LINKED ISSUE #198: Cards have scroll bars on them which looks bad

AARON-CLARK:
Seeks to close #198


Jeff-Thompson12 (changes requested):
Looks really good. Only thing I noticed is that downloading HTML reports on the "Report Preview" tab doesn't work. The HTML reports are still working if requested from the "Database" tab though.

AARON-CLARK:
> I noticed is that downloading HTML reports on the "Report Preview" tab doesn't work. 

hmmmmm, sees to be working on my side @Jeff-Thompson12! Could you try restarting your Rstudio session by chance?

Jeff-Thompson12:
I closed all instances of RStudio. Still getting the same issue. It can't find `auto_font()`. Can you try restarting your R environment and running `golem::run_dev()` again?

image

AARON-CLARK:
@Jeff-Thompson12, ohhhhh, whoops!  Apologies. It should be fixed now.

Jeff-Thompson12:
@AARON-CLARK I didn't notice this previously, but the sizing appears to be off for the HTML report downloads. It appears that the column in the app is wider than the downloaded report. This results in a section header like "Maintainer" going from one line to two and pushing text outside of the card.

image

AARON-CLARK:
> It appears that the column in the app is wider than the downloaded report. This results in a section header like "Maintainer" going from one line to two and pushing text outside of the card.

@Jeff-Thompson12, I could certainly make the title font-size smaller for the cards in `reportHTML.Rmd` but I fear this use case with Thomas' long email address will still push the card's footer around. I could leverage an ellipsis `...` to truncate long maintainer names/email addresses but it would be sad since we'd lose some precious info that's actually usable (like a full email address). It's was okay to trim the URLs because they have an `href` attached to them, making them clickable.

To be honest, I'm not entirely sure why sometimes the footer get's jostled from long text and sometimes it doesn't.

AARON-CLARK:
Hi @Jeff-Thompson12, scratch that. I just adjusted the margins of the entire html report and that seemed to widen things out, creating more space (commit b6def26).

inst/app/www/reportHtml.Rmd
```diff
@@ -25,6 +25,14 @@ params:
   downloads_plot_data: NA
 ---
 
+<style type="text/css">
+.main-container {
+  max-width: 1200px;
```
Jeff-Thompson12, line 30: Might be a bit too nit-picky, but maybe something like `max-width: min(1200px, 80%);` so that margins still exist on small screens? What do you think?

Jeff-Thompson12 (commented):
Looking pretty good. I like changing the width of the HTML report itself. The margins felt too wide anyways. The report did seems a bit wide on my laptop (see below), but should be an easy adjustment.

AARON-CLARK:
@Jeff-Thompson12, it is done. Thanks!

Jeff-Thompson12 (approved):
Everything LGTM!

PR #297, opened 2022-11-22
new run_app args for pre_auth_user

Robert-Krajcik:
Another interpretation of issue #278. using server-side coding.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: riskassessment
 Title: A web app designed to interface with the `riskmetric` package
-Version: 0.0.1.9006
+Version: 0.0.1.9007
 Authors@R: c( 
     person("Aaron", "Clark", role = c("aut", "cre"), email = "aaron.clark@biogen.com"),
     person("Robert", "Krajcik", role = "aut", email = "robert.krajcik@biogen.com"),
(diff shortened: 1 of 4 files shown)


Jeff-Thompson12:
@Robert-Krajcik It doesn't always work for me. All I can figure is that sometimes the `updateInputText()` fires before the UI elements exist.

Also has the same problem as my PR where it fires when you log out. The only way to switch users is to change `run_dev.R` or run `run_app()` instead of `golem::run_dev()`.

Robert-Krajcik:
Thanks for letting me know, @Jeff-Thompson12 
I had the debounce parameter down to 500, maybe I should bring that back up.
And yes, when you log out, it blithely logs you back in.
And it's waaayyy over-coded.  No need to search through the credentials db 
if this is intended for developers with admin authority.

Jeff-Thompson12:
I don't think the `debounce()` parameter is the issue. When it does not work, the UI elements do not get updated, so the `updateInputText()` is not firing at the right time.

Robert-Krajcik:
With this, a user would have to supply
`options(pre_auth_user = TRUE)`
and credentials somewhere (on the console, perhaps?)
Example user_id and password below.
`options(login_creds = list(user_id = "UberAdmin", user_pwd = "ASDfgh1"))`

Robert-Krajcik:
redundant with #293

Issue #280, opened 2022-11-16
Create PDF report export type


AARON-CLARK:
Right now, the app supports HTML & DOCX file types.

Robert-Krajcik:
Closed with PR #300 Add pdf report option

PR #300, opened 2022-11-29
Add pdf report option

Robert-Krajcik:
Addresses Issue #280

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # riskassessment (development version)
 
-* Added new pdf report download option to 'Database' and 'Report Preview' tabs. 
+* Added new pdf report download option to 'Database' and 'Report Preview' tabs (#280) 
 * Updated README to include more description, with a focus on improving the 'Usage' section, which now shows our audience how to install & run the app for the first time. Also, new links to the demo app & a short video walk through were added. Plus, very brief notes were included regarding deployment environments.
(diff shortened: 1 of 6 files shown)


NEWS.md
```diff
@@ -1,5 +1,6 @@
 # riskassessment (development version)
 
+* Add pdf option to databaseView and reportPreview
```
AARON-CLARK, line 3: Maybe be a little more description would help here. Consider amending so it makes more sense for the reader since users won't know those shiny module names. Also, if you could please link to the github issue, that'd be helpful. For example:
>Added new pdf report download option to 'Database' and 'Report Preview' tabs (#300).

AARON-CLARK (changes requested):
Hi @Robert-Krajcik, thanks for the PR, it works great! I have a couple of comments:

* Could you please run `usethis::use_version("dev")` to increment the dev version number?

* I just got a new laptop, and needed to install the `{tinytex}` package for this to work, so perhaps you could run `usethis::use_package("tinytex", type = "Suggests")` to add that package to the `Suggests` field.

Thanks!

NEWS.md
Robert-Krajcik: OK, incremented version number and added `{tinytex}` to suggests field, and updated NEWS.md

AARON-CLARK (approved):
LGTM!

Issue #295, opened 2022-11-21
Design unit tests for each function


AARON-CLARK:
Hi all, I took the time to group our functions into similar groups. The hope is that we can all contribute to writing `testthat` tests. Please review some of the categories below and feel free to volunteer for a certain group/ category of test. I submit a PR (#302) that will actually re-organize some of these functions to new utils files so they are easier to track down in the `R/` folder. Please stay tuned until it is merged into `dev`, thank you!

|Group     |Function                            |Developer |PR submitted | PR Reviewed |
|:---------|:--------------------------------|:---------|:---------|:---------|
|app            |run_app                          |          |         |         |
|get db         |dbSelect                         |    Robert-Krajcik      |   X    |    X    |
|get db         |get_cm_comments                  |  Robert-Krajcik        |   X    |    X    |
|get db         |get_comm_data                    |   Robert-Krajcik       |   X    |    X    |
|get db         |get_metric_weights               |   Robert-Krajcik       |   X    |    X    |
|get db         |get_mm_comments                  |   Robert-Krajcik       |   X    |    X    |
|get db         |get_mm_data                      |   Robert-Krajcik       |   X    |    X    |
|get db         |get_overall_comments             |   Robert-Krajcik       |   X    |    X    |
|get db         |get_pkg_info                     |  Robert-Krajcik        |   X    |    X    |
|get db         |weight_risk_comment              |  Robert-Krajcik        |   X    |    X    |
|insert db      |dbUpdate                         |  Robert-Krajcik        |    X   |    X  |
|insert db      |insert_community_metrics_to_db   |  Robert-Krajcik        |    X    |   X    |
|insert db      |insert_maintenance_metrics_to_db |  Robert-Krajcik        |    X   |    X    |
|insert db      |generate_comm_data            | Aaron Clark |   X    |     X    |
|insert db      |insert_pkg_info_to_db            |  Robert-Krajcik        |   X     |   X     |
|insert db      |update_metric_weight             |  Robert-Krajcik        |    X     |   X     |
|insert db      |upload_package_to_db             |  Robert-Krajcik        |    X    |    X    |
|metric modules |addCommentServer                 |borgmaan|         |         |
|metric modules |addCommentUI                     |borgmaan|         |         |
|metric modules |communityMetricsServer           |    Robert-Krajcik                |   X    |   X   |
|metric modules |communityMetricsUI               |    Robert-Krajcik                |   X   |    X   |
[...]

PR #303, opened 2022-12-05
Example `testthat` on modules

AARON-CLARK:
This PR is a test to explore what testing modules might look like.

Partially solves #295.


AARON-CLARK:
Stale - we have adopted a `shinytest2` approach which is more robust.

PR #306, opened 2022-12-06
Add `fontawesome != 0.3.0` warning

### LINKED ISSUE #275: Report download not working with `{fontawesome}` Version `0.4.0`

AARON-CLARK:
closes #275


AARON-CLARK:
Converting to draft because:

image

Jeff-Thompson12:
@AARON-CLARK I think you can only have minimum version requirements.

Perhaps we should check the version of `{fontawesome}` when creating the HTML report and create a helpful message.

AARON-CLARK:
> @AARON-CLARK I think you can only have minimum version requirements.

I've set a version using `==` before and that plays nicely when building/installing the package, etc. However, it looks like the `r-lib/actions/setup-r-dependencies@v2` portion of our GHA workflow is struggling with it.

> Perhaps we should check the version of `{fontawesome}` when creating the HTML report and create a helpful message.

I'm okay with that approach too! I'll draft something up real quick.

AARON-CLARK:
New Login note:
image

R/run_app.R
```diff
@@ -34,9 +34,17 @@ run_app <- function(
...
+  if(is.null(login_note)) {
+    # TODO: Remove temporary warning once bug in fa v0.4.0 is fixed.
+    # https://github.com/rstudio/fontawesome/issues/99
+    # Here, we make sure user has a functional version of fontawesome
+    fa_v <- packageVersion("fontawesome") #TODO: Remove once bug is fixed
+    if(!file.exists(credentials_db_name)) {
+      login_note <- HTML('<em>Note:</em> To log in for the first time, use the admin user:
+                          <u>admin</u> with password <u>QWERTY1</u>.')
+    } else if(fa_v != '0.3.0') { #TODO: Remove once bug is fixed
+      login_note <- HTML(glue::glue("<em>Note:</em> HTML reports may require fontawesome 0.3.0 to render. You currently have v{fa_v} installed. If the report download fails, please run: `remotes::install_version('fontawesome', version = '0.3.0', repos = 'http://cran.us.r-project.org')`"))
```
Jeff-Thompson12, line 46: "fontawesome v0.3.0" to be consistent with other text.

dev/02_dev.R
```diff
@@ -183,6 +184,10 @@ usethis::use_data_raw( name = "adsl", open = FALSE ) # not run.
 ## Tests ---- not run
 ## Add one line by test you want to create
 usethis::use_test( "app" )
+usethis::use_test("assessmentInfoServer")
+
+golem::add_module( name = "big_fake" ) # ran
+golem::use_module_test("big_fake")
 
 # Documentation
 ## Vignette ---- notrun
```
Jeff-Thompson12, line 193: Looks like a commit from another PR snuck it's way in here.

Jeff-Thompson12 (commented):
Do we also want to print the warning to the console in databaseView.R and reportPreview.R? Might be a bit overkill, but is pretty low hanging fruit and the processing for the Rmd file is being printed to the console. Might be good to have the warning show up there as well.

dev/02_dev.R
AARON-CLARK: Whoops!

R/run_app.R
AARON-CLARK: done.

AARON-CLARK:
> Do we also want to print the warning to the console in databaseView.R and reportPreview.R?

It already was!
image

dev/02_dev.R
Jeff-Thompson12: There is also the line `usethis::use_package("rjson", type = "Suggests")

dev/02_dev.R
```diff
@@ -65,6 +65,7 @@ usethis::use_mit_license()
 # usethis::use_package("glue")
 # usethis::use_package("lazyeval",type="Suggests")
 # usethis::use_package("scales",type = "Suggests")
+usethis::use_package("rjson",type = "Suggests")
```
Jeff-Thompson12, line 68: Can this also be removed?

Jeff-Thompson12 (changes requested):
It's a bit nit picky, but doesn't make sense to have a line for `{rjson}` in dev when it's not being used in the package. Or at least not yet.

dev/02_dev.R
AARON-CLARK: removed.

Jeff-Thompson12 (approved):
Looks great.

PR #310, opened 2022-12-07
create some testthat tests for database functions

Robert-Krajcik:


Robert-Krajcik:
So copying a "skeleton" of the `database.sqlite db` into `./inst/testdata` worked.
Both with `devtools:::check()` and the auto-checks above

Robert-Krajcik:
There are more tests to do, but let's start with this one

Robert-Krajcik:
Right now, I am extensively testing `dbSelect()` and `dbUpdate()` functions.
For the latter, I copy the `skeleton.sqlite` db into a "temporary" db (not ":memory:" as I need to open and close it a few times)
and then `unlink()` the temp db when I am done.

.gitignore
```diff
@@ -1,5 +1,5 @@
 rsconnect/
-*.sqlite
+.sqlite
```
Jeff-Thompson12, line 2: Please adjust this to `/*.sqlite` as seen [here](https://github.com/pharmaR/risk_assessment/pull/312/files#diff-bc37d034bad564583790a46f19d807abfe519c5671395fd494d8cce506c42947L2-R2). This tells git to restrict the ignore to the main directory and not subdirectories. That way the credentials, credentials_dev, and database SQLite files are not being added to the tracking. For reference: https://git-scm.com/docs/gitignore.

Jeff-Thompson12 (changes requested):
@Robert-Krajcik I like the sequential nature of the tests. One thing I do think is missing is verifying that the error catching is working. Both `dbSelect()` and `dbUpdate()` have a `tryCatch()`. I think we should also verify that bad queries/commands will be caught.

.gitignore
Robert-Krajcik: OK on the changes above

tests/testthat/test-dbSelect.R
```diff
@@ -0,0 +1,36 @@
...
+  base_path <- app_sys("testdata")
+  
+  # 1. file exists?
+  testthat::expect_true(file.exists(file.path(base_path, db_name)))
+  
+  # 2. valid db? test this is a valid sqlite database
+  testthat::expect_equal(readLines(file.path(base_path, db_name), n =1, warn = FALSE), "SQLite format 3")  
+
+  # 3. can we connect?
+  testthat::expect_true(DBI::dbCanConnect(RSQLite::SQLite(), db_name))
```
Jeff-Thompson12, line 15: So I noticed that this when ran locally actually creates a skeleton.sqlite file in my testthat folder because you have `db_name` instead of `file.path(base_path, db_name)`. This is also interesting that no error is raised despite the SQLite database not existing.

Robert-Krajcik: I am going to fix that

.gitignore
Jeff-Thompson12: For some reason those files are still being tracked...

Jeff-Thompson12: Appears you actually need to remove it from the cache: https://stackoverflow.com/questions/1274057/how-do-i-make-git-forget-about-a-file-that-was-tracked-but-is-now-in-gitignore#:~:text=on%20this%20post.-,.,that%20are%20already%20being%20tracked.&text=The%20removal%20of%20the%20file,happen%20on%20the%20next%20commit.

Evidently `.gitignore` only applies to files that are not being tracked.

Robert-Krajcik: OK done.
I ran `git rm -r --cached . && git add . && git commit -am "Remove ignored files"`
and pushed the change

Robert-Krajcik:
Ready for another review, @Jeff-Thompson12

NEWS.md
```diff
@@ -1,5 +1,6 @@
 # riskassessment (development version)
 
+* Added testthat tests for dbSelect and dbUpdate functions (#295)
```
Jeff-Thompson12, line 3: This feels super nitpicky on my part but we have been adding the NEWS comments at the bottom of this section. (i.e. line 17)

Jeff-Thompson12 (changes requested):
Can you also update the Version # to 0.0.1.9012 in the DESCRIPTION file?

Jeff-Thompson12:
@Robert-Krajcik Everything looks really good! Just a couple of really small adjustments.

NEWS.md
Robert-Krajcik: OK on the above

Jeff-Thompson12 (approved):
LGTM!

PR #315, opened 2022-12-19
48 make app more portable with renv

Xyarz:
added the renv with the lock file as well as a small Rmd to explain the usage of renv.


Robert-Krajcik (changes requested):
Hi Lars

First thing: edit the DESCRIPTION file and add a comma in the authors@R field after this line
`person("Munshi Imran", "Hossain", role = "ctb")`
This is causing a DESCRIPTION read failure while running devtools::check()

Next please run golem::document_and_reload()
to update `man/riskassessment-package.Rd`

Thirdly, please add an entry to NEWS.md
Thanks,
Robet

Xyarz:
There seems to be an issue with having both renv and pak.
It seems as if they just recently [updated](https://github.com/r-lib/actions/blob/v2-branch/setup-r-dependencies/action.yaml) the github action workflow in order to manage that.

Robert-Krajcik:
Hi @Xyarz 
I see the `devtools::check()` works now, but the github actions are all failing.
This needs to get resolved before I can approve this.

Also note that the renv.Rmd file is not rendering
image

One way to fix this:
```{r setup, echo= FALSE}
pkgload::load_all(quiet = TRUE, warn_conflicts = FALSE, attach_testthat = FALSE)
library(riskassessment)
```
Another might be to simply remove the `library()` function

borgmaan:
> Hi @Xyarz,
> 
> @borgmaan has initiated `renv` and it's passing `devtools::check()` on #322. Do you want to try to resolve the issue (the renv/pak conflict) you referenced on this branch , or just proceed forward with Andrew's branch? Please take a look and let us know your thoughts.

I would propose these two be kept separate and work continue over here to properly set up `renv` there is some nuance to doing this right and my goal with what was done on [`ab-add-shinytest2`](https://github.com/pharmaR/risk_assessment/tree/ab-add-shinytest2) was to quickly allow others to reproduce and confirm it was not a package compatibility issue. I plan to delete that `renv` prior to merging.

AARON-CLARK:
At 1/31 meetup, @Xyarz said he'll prioritize this PR this week and the only remaining issue is related to libcurl. Thanks! I think this will be important for our upcoming release.

Robert-Krajcik:
I'm looking this over now @borgmaan, but would like another set of eyes on this.  Thanks.

renv.lock
```diff
@@ -0,0 +1,2325 @@
+{
+  "R": {
+    "Version": "4.1.2",
+    "Repositories": [
+      {
+        "Name": "CRAN",
+        "URL": "https://cran.rstudio.com"
```
borgmaan, line 7: I propose pointing to a stable snapshot (either from MRAN or public RSPM). This will improve stability and prevent developers from always being prompted to upgrade packages. 

See [here](https://github.com/RConsortium/submissions-pilot2/blob/main/renv.lock#L7) for an example of this approach being used in the R submissions working group.

vignettes/renv.Rmd
```diff
@@ -0,0 +1,47 @@
...
+
+First one has to initialize a new project-local environment with a private R library - which can be done via `renv::init()`.
+Afterwards, work in the project as usual and save the current state of the project library with `renv::snapshot()`.
+This result will be saved into a lockfile - named `renv.lock`.
+Now, continue working on your project and call `renv::snapshot()` again, in order to update the project state - or with `renv::restore()`, revert the previous state if you have encountered problems while developing.
+
+## How to ignore files in the lock file
+
+By default `renv` ignores all files stated in the `.gitignore` file. If you want to extend that further, you can use a `.renvignore` file - which is specified similar to the `.gitignore` file.
+
```
borgmaan, line 38: It may be worth adding logic here around the dev team philosophy of upgrading or changing the existing `lockfile` once it is stable. 

My suggestion would be to never upgrade once stable unless there is a very compelling reason to do so (e.g. new feature we can't live without).

borgmaan (changes requested):
@Xyarz -- there is still some work to be done to get the GHA workflows working as expected with `renv`. 

All workflows should be updated to use `r-lib/actions/setup-renv` in place of `r-lib/actions/setup-r-dependencies`. Some of the workflows are currently using both (duplicative) and others are only using `r-lib/actions/setup-r-dependencies`. 

`r-lib/actions/setup-r-dependencies` uses `pak` under the hood and does not respect the `renv` lockfile. It includes it's [own logic](https://github.com/r-lib/actions/blob/v2/setup-r-dependencies/action.yaml#L78-L) for dependency resolution and can install versions that are mismatched with what is included in `renv.lock` leading to unexpected behavior. 

I dropped a couple of other comments regarding the repo source and philosophy on changing the lockfile once stable that may warrant feedback from the broader team (@AARON-CLARK ).

Robert-Krajcik (changes requested):
I admit I'm new to working with `renv`
But I am getting this note:
```{r, eval=FALSE}
The following package(s) are used in the project, but are not installed:

	checkhelper
	revdepcheck
	riskassessment

Consider installing these packages, and then using `renv::snapshot()`
to record these packages in the lockfile.
```
When I go to install `revdepcheck`
```{r, eval=FALSE}
> install.packages("revdepcheck")
Error: package 'revdepcheck' is not available
In addition: Warning messages:
1: package 'revdepcheck' is not available 
2: package 'revdepcheck' is not available from MRAN
```
Can you look into this?

renv.lock
Xyarz: good idea thanks!

vignettes/renv.Rmd
AARON-CLARK: That sounds good to me. @Xyarz, could you add a sentence or two that summarizes what Andrew has stated above?

Xyarz: added in [a5a96fa](https://github.com/pharmaR/riskassessment/pull/315/commits/a5a96fa7644de7e69f087394effc1ab045739b32)

AARON-CLARK (approved):
LGTM

PR #317, opened 2022-12-20
added tests for getTimeStamp function

Xyarz:


Robert-Krajcik (dismissed):
Hi Lars,
Again, I would replace the top-level `describe` with `test_that` so the 'run tests' command appears 
And I would keep the `getTimeStamp()` and `Sys.time()` functions as close  together as possible
```{r, eval=FALSE}
      expect_equal(object = getTimeStamp(),
                   expected = paste(gsub(x = Sys.time(), pattern = " ", replacement = "; "),
                                    Sys.timezone())
```
Other than those two things, it looks good.

AARON-CLARK (approved):
LGTM

PR #318, opened 2022-12-20
added tests for get_latest_pkg_info

Xyarz:


Robert-Krajcik (dismissed):
Hi Lars
The top-level `describe` needs to be replace with a `test_that`

Personally, I would just drop the `describe` and `it` blocks altogether 
and just run 
`test <- get_latest_pkg_info("rpact")` once
and then have the 3 `expect_*` functions follow it.

AARON-CLARK (approved):
Couldn't have done better myself! Pleased to finally see this one get through!

PR #323, opened 2023-01-03
Rk 281 dependencies

Robert-Krajcik:
Shrinking the list of package dependencies...
So far, I have eliminated `{magrittr}`, `{waiter}` and `{tibble}`
`Imports includes 30 non-default packages.`
Still looking to see if there are any others.


Robert-Krajcik:
So far I have it down from 33 to 27 non-default packages.
```
checking package dependencies ... NOTE
  Imports includes 27 non-default packages.
  Importing from so many packages makes the package vulnerable to any of
  them becoming unavailable.  Move as many as possible to Suggests and
  use conditionally.
```

AARON-CLARK:
> * Can we live without `shinydashboard::box`

I don't think so. I kind of looked into it, and it's an important piece of all our cards in the app. Maybe it'd be worth seeing if there is something similar in `shiny`?

AARON-CLARK:
> * Can we move `rmarkdown` to suggests?

Unfortunately no, because we used `rmarkdown::render()` to create our downloadable reports.

AARON-CLARK:
> * Can we replace `glue` with `paste|paste0`?

That's certainly an option. I know `glue` is used pervasively throughout the package and really simplifies a lot of syntax for queries. Maybe keep it for now, and if we get desperate, we can hack it up later?

Robert-Krajcik:
Thanks for looking at this, @AARON-CLARK !
I could look into replacing one instance of `pivot_wider` with `reshape`
Other than that, I don't see any other packages we could easily remove.

Robert-Krajcik:
Hi @AARON-CLARK 
Now we are down to 26 packages, from 33 
`Imports includes 26 non-default packages.`

AARON-CLARK (changes requested):
Hi @Robert-Krajcik, this looks really good! Thanks for drafting this PR. I don't see any issues with the changes & clean up you've performed. If you increment the package to be one dev version higher than the [current](https://github.com/pharmaR/riskassessment/blob/eedc8376a3868727a7d44b20b2fe3deb60b48461/DESCRIPTION#L3) `dev` branch & add a bullet point to the `NEWS.md` file, I'll be glad to merge.

AARON-CLARK (approved):

Issue #324, opened 2023-01-04
Color Palette not savvy for colorblindess


AARON-CLARK:
We've received some feedback indicating its difficult for colorblind folks to tell what colors are good/average/bad in the app based on the colors shown below. That is, they weren't convinced the darkgreen was better... which is problematic. This is what our demo app's database has always looked like:

![image](https://user-images.githubusercontent.com/17878953/210579344-507248e1-6520-4284-ac44-c735e5e20ba9.png)

AARON-CLARK:
And this is what it looks like with a possible solution, using color palette: `viridisLite::turbo(11, begin = 0.4396875, end = .8728125)`

![image](https://user-images.githubusercontent.com/17878953/210613754-049f5722-0c6d-485e-84ec-79c742b4f76e.png)

AARON-CLARK:
closed with #235

PR #325, opened 2023-01-04
Fixes issue for colorblind users

AARON-CLARK:
See examples in #324 for the 'before image' but the database tab looks like the below image with this changes in this PR. Notice, I had to dynamically change the color of the font in some areas. Overall, I tend to like the old color gradient better, but this is confirmed to work with our colorblind users.

image


Jeff-Thompson12:
@AARON-CLARK This is probably the request you were dreading to hear, but can we make this a toggleable option?

AARON-CLARK:
> @AARON-CLARK This is probably the request you were dreading to hear, but can we make this a toggleable option?

@Jeff-Thompson12, I suppose it could be. But why? Because we liked the other color scheme better? I'm not opposed.

Jeff-Thompson12:
Part of me wonders why not go the whole way and let users select their own palette, but I suppose simpler is better. I guess there is no compelling reason to toggle an "accessibility" view.

AARON-CLARK:
@Jeff-Thompson12 mentioned it may be nice to have other visual indicators besides colors, like an exclamation point or a thumbs up or emoji faces next to the each decision.

AARON-CLARK:
So, good news, I'm told that if you darken a colorblind-friendly palette, they can still perceive the gradient. So I think this is a happy compromise and doesn't look oppressively bright. Thoughts @Jeff-Thompson12?

image

Jeff-Thompson12 (approved):
I like the more muted colors. I still think it could be nice to have some form a visual indicators outside of color, but that feels like a different PR/issue.

Issue #248, opened 2022-06-14
Ability to type in package names


thebioengineer:
Currently the app expects users to upload a csv with package and version. This takes users outside the app and adds friction. It would be nice to have the ability to write a package name and version directly in the UI of the app.

AARON-CLARK:
Duplicate of #247, but we'll keep open since #247 contains multiple issues.

Thanks for the feedback @thebioengineer! We were already starting to make mental preparations for this upgrade (at least for CRAN packages). What do you think of the UI experience on [diffify](https://diffify.com/R/shiny/1.7.0/1.7.1)? Does it fulfill your needs from an ease-of-use perspective?

AARON-CLARK:
closed with #327 and #329

PR #327, opened 2023-01-05
Type Package Names

Jeff-Thompson12:
Addresses #248

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: riskassessment
 Title: A web app designed to interface with the `riskmetric` package
-Version: 0.0.1.9012
+Version: 0.0.1.9013
 Authors@R: c( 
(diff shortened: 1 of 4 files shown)


Jeff-Thompson12:
I did not incorporate adding a package version since that is not being taken into account in the code base. If/when that functionality is added, we can address adding a version to the selectize input.

Robert-Krajcik (dismissed):
This looks good to me, @Jeff-Thompson12 
I see you updated the rintrojs (help) and it looks good.
And I can still enter a .csv file full of non-existent package names and that still works.

Jeff-Thompson12:
@Robert-Krajcik Great catch! It doesn't look like we were checking package names from the csv either (outside of trimming the white space). I added in some code to keep `riskmetric` from erroring out.

Robert-Krajcik (approved):
OK I am no longer able to break it.  Good job!

PR #329, opened 2023-01-06
A few 'Type Package Name' edits

AARON-CLARK:
Make pre-load `actionLink` disappear after clicked and re-arrange ui so that package `selectizeInput` comes first!

image


AARON-CLARK:
Hey @Jeff-Thompson12, I read over #327 and saw you didn't automatically 'Pre-load all CRAN packages' because it took an half a second to load up, but I'm wondering if there is some way we could avoid having an `actionLink()`? Any ideas?

Jeff-Thompson12:
@AARON-CLARK One thought I had while creating the PR was to create an observer that would run if the selectize input was selected. I can experiment with adding that type of functionality to this PR/branch.

Jeff-Thompson12:
@AARON-CLARK Here is a [patch.txt](https://github.com/pharmaR/riskassessment/files/10376007/patch.txt) built off of this branch. It adds a function to trigger `input$load_cran` when focus is on the selectize input. I doubt the user will even notice that's when the loading of the input options is occurring.

Robert-Krajcik (changes requested):
I am starting with a clean database.sqlite here and I don't see the
pre-load action link at all.  

image

Jeff-Thompson12:
> I am starting with a clean database.sqlite here and I don't see the
> pre-load action link at all.

@Robert-Krajcik That is by design at request by Aaron. The list will now get loaded into memory when focus is applied to the selectize input. If you start typing packages in the text box, you should see the CRAN packages pop up after a brief delay.

Jeff-Thompson12:
I also just pushed a commit that will adjust the position of the button based on scroll height of the selectize input's label. This ensures that the button will be in the appropriate spot in a smaller window.

AARON-CLARK:
@Jeff-Thompson12, great work with commit [0c47f76]!(https://github.com/pharmaR/riskassessment/pull/329/commits/0c47f76d8ccdf768fe68f6f83f3ea3039c1397d8)

Jeff, or @Robert-Krajcik, I'd say this PR is ready to go. Please merge when ready.

PR #331, opened 2023-01-06
updated readme links with new repo name

AARON-CLARK:


Jeff-Thompson12:
Yes, I noticed the redirect on accident. I missed the public service announcement that the URL had changed, but it wasn't affecting my ability to push or pull.

AARON-CLARK:
> I missed the public service announcement that the URL had changed

I announced it at our Thursday meeting that I'd do it Friday. The official PSA was on slack! Do you think we need a public facing PSA?

Jeff-Thompson12:
> Do you think we need a public facing PSA?

It might be something worth talking about. We did not actually change the name of anything in some sense and it looks like the [redirect is intended behavior](https://docs.github.com/en/enterprise-cloud@latest/repositories/creating-and-managing-repositories/renaming-a-repository) too.

dev/run_dev.R
```diff
@@ -12,7 +12,7 @@ golem::detach_all_attached()
 golem::document_and_reload()
 
 # Run the application 
-run_app(pre_auth_user = TRUE)
+run_app()
```
Jeff-Thompson12, line 15: Why did you remove the `pre_auth_user = TRUE` option?

docs/Managing_Userids_and_Passwords.Rmd
```diff
@@ -208,7 +208,7 @@ next log-in.
 
 ## Have an issue?
 
-That's all for the tutorial. If you have any issues or feature requests as it pertains to managing user IDs and passwords in the app, please open a new issue on our [Github repo](https://github.com/pharmaR/risk_assessment/issues). Merry managing!
+That's all for the tutorial. If you have any issues or feature requests as it pertains to managing user IDs and passwords in the app, please open a new issue on our [Github repo](https://github.com/pharmaR/riskassessment/issues). Merry managing!
```
Jeff-Thompson12, line 211: The Managing_Userids_and_Passwords.html file was not updated with this change.

Jeff-Thompson12 (changes requested):
All the URLs work. I see that we have to wait to update the URL for the demo. Just a couple things:

- [ ] Can we update Using_SQLite_Command_Line.Rmd and Using_SQLite_Command_Line.html as well?
- [ ] Can we update the R project file from risk_assessment.Rproj to riskassessment.Rproj?

Robert-Krajcik (commented):
Both the demo and the "quick tour" by Marly Cormar are working.

docs/Managing_Userids_and_Passwords.Rmd
AARON-CLARK: @Jeff-Thompson12 fixed.

dev/run_dev.R
AARON-CLARK: @Jeff-Thompson12 I'm actually not sure. I might have been testing something? I've returned in a recent commit.

AARON-CLARK:
> * Can we update Using_SQLite_Command_Line.Rmd and Using_SQLite_Command_Line.html as well?

@Jeff-Thompson12: the repo link doesn't exist in this Rmd/html file. Or are you referring to the directory references [here ](https://github.com/pharmaR/riskassessment/blob/1f6b12032be45904ecef4dbc32d441be66f3c3a5/docs/Using_SQLite_Command_Line.Rmd#L93) and [here](https://github.com/pharmaR/riskassessment/blob/1f6b12032be45904ecef4dbc32d441be66f3c3a5/docs/Using_SQLite_Command_Line.Rmd#L152)? This brings up a good point that maybe should be addressed in this vignette: we have no control over what people name their directories, so maybe the vignette should mention "we'll assume the directory's name is 'riskassessment'" for most who clone the repo. I'll pushed a commit 66c95d8 to add this.

AARON-CLARK:
> Some suggested wording changes to `README.Rmd'

Thanks @Robert-Krajcik. Added!

Jeff-Thompson12 (approved):
LGTM! @Robert-Krajcik did you have anything else?

PR #343, opened 2023-01-13
Testing dependency fix for databaseView

borgmaan:


Jeff-Thompson12:
@borgmaan I see. The download handler failed on GHA because of font awesome version 0.4.0. Well I'm going to steal that bit of code for my PR. Thanks!

PR #335, opened 2023-01-11
Rk 147 user guide

Robert-Krajcik:
Adding a user guide and a vignettes folder

Issue #147, opened 2021-12-03
Create & publish user guide


AARON-CLARK:
Since we are making an official release, we need documentation to show people how to use the app! Somewhat related to #133. Unfortunately, risk_assessment isn't a package, so I presume we can't use `pkgdown` (Whomp whomp).

AARON-CLARK:
Just brainstorming - we'll have to introduce these topics to users:

- [ ] Logging in
- [ ] The pkg upload tab
- [ ] The sidebar
- [ ] Maintenance metrics
- [ ] Community Usage Metrics
- [ ] The pkg database screen (including admin features)

Since there are 6, maybe we could split up and each write two vignettes each?

Robert-Krajcik:
I could work on logging in (and setting up userids)
and the pkg upload tab

marlycormar:
This brings a good point @Robert-Krajcik. We have app users and app admins (and app developers), with different interests in the app. Should we split the documentation?
For instance, admins may want to know how to set up the app for the first time; users will not be involved in setting up the app, but are interested in running reports, making final decisions, etc; developers may need to understand the app's structure (code), which is related to #58.

Robert-Krajcik:
Yeah the users wouldn't need to know how to create and modify userids and passwords.
So that would be a separate vignette for admins.

Robert-Krajcik:
I am thinking I can put together a guide for non-admin users,
but the pkg database screen (with admin features) would probably 
need to be a separate vignette.  Thoughts??

AARON-CLARK:
> the pkg database screen (with admin features) would probably
> need to be a separate vignette. Thoughts??

Agreed. Separate vignette for admin features.

Robert-Krajcik:
So far I have created #152 for admins on managing user ids and passwords.
I am going to create another for admins on Changing Risk Score Weights by Metric.
And then perhaps someone could write a general user guide for everyone?

marlycormar:
Documentation:
- [ ] For admins: Managing user ids and passwords (PR open) @Robert-Krajcik 
- [ ] For admins: Managing risk score weights (Needs PR) @AARON-CLARK 
- [ ] For users: How to use the app?  @Robert-Krajcik
      - [ ] Add small video that shows an example using the app (look for video) @marlycormar 
      - [ ] Explain each tab of the app: include how to create a report.
      - [ ] Explain the apps philosophy: how to upload a package? Talk about the final decision.
[...]

PR #349, opened 2023-01-13
Augment Robert's User guide + add branding images

AARON-CLARK:
Built upon #335 (Thanks @Robert) and makes progress on #147.

When you knit the `riskassessment.Rmd` file, don't worry too much about how wide certain images are in the vignette, those will automatically get scaled down per `pkgdown`.

I also updated the `README` with some sexy `riskmetric` and `riskassessment` hex stickers, but I also changed the favicon and added some imagery in-app. Here are some snapshots...

image

image

image

image

I would love to solicit someone's help to get that last one up into the top-level `tabsetPanel()`. Or thoughts on if we even need that image? Perhaps something on the auth page would suffice?


Robert-Krajcik (changes requested):
Well you have now covered every square inch of this package, @AARON-CLARK 

My only requested change is about the 'Administrative tools` tab.
There are two sub-tabs on this page.
One is the `Credential Manager` while the other is `Assessment Reweighting`

You describe the Credential Manager tab while showing the Assessment Reweighting.

I think you should describe both tabs here and have snaphots of each one.

I assume we can go back and put in the {"insert links here"] at a later time.

AARON-CLARK:
Thanks for the review @Robert-Krajcik. I added your suggestion in commit 7335df6.

image

I'll convert this to draft while I figure out why R CMD Check was failing.

AARON-CLARK:
Okay @Robert-Krajcik, she should be ready for review again. One of the vignette pngs was "non-portable" since it had a space in the name. So, the R CMD Check should not fail.

Robert-Krajcik (dismissed):
Looks great @AARON-CLARK

Robert-Krajcik (approved):
Looks good to me

PR #350, opened 2023-01-16
Tests for introJS[UI/Server]

Jeff-Thompson12:
Adds tests for the `introJS` module. Adjustments to the app code has been made where necessary to enable more full testing.


Robert-Krajcik (approved):
After installing riskmetric and setting
`options(shinytest2.load_timeout = 30*1000)`
All tests passed.

PR #355, opened 2023-01-17
Adjust GHAs for PRs with `dev` base branch.

AARON-CLARK:
Does this change look alright? I don't think we need to increment the dev version or include this in the NEWS file.

### DIFF
--- a/.github/workflows/R-CMD-check-dev.yaml
+++ b/.github/workflows/R-CMD-check-dev.yaml
@@ -20,9 +20,9 @@ jobs:
         config:
           # - {os: macOS-latest,   r: 'release'}
           # - {os: windows-latest, r: 'release'}
-          - {os: ubuntu-latest,   r: 'devel', http-user-agent: 'release'}
+          # - {os: ubuntu-latest,   r: 'devel', http-user-agent: 'release'}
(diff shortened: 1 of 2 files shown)


.github/workflows/R-CMD-check-dev.yaml
```diff
@@ -0,0 +1,50 @@
...
+    name: ${{ matrix.config.os }} (${{ matrix.config.r }})
+
+    strategy:
+      fail-fast: false
+      matrix:
+        config:
+          # - {os: macOS-latest,   r: 'release'}
+          # - {os: windows-latest, r: 'release'}
+          - {os: ubuntu-latest,   r: 'devel', http-user-agent: 'release'}
+          - {os: ubuntu-latest,   r: 'release'}
```
borgmaan, line 24: I would propose only running on `{os: ubuntu-latest, r: 'release'}` for now.

borgmaan (dismissed):

.github/workflows/R-CMD-check-dev.yaml
AARON-CLARK: resolved with commit 5835054!

Issue #354, opened 2023-01-17
Some `shinytest2` tests causing inconsistent test failures


AARON-CLARK:
![image](https://user-images.githubusercontent.com/17878953/212944996-928caca1-c460-47f3-9425-56b1de5e0531.png)

Effects these PRs right now:
* #349 
* #352
* #318 
* #317

AARON-CLARK:
Problem still exists, but annoyance is reduced by reducing matrix of R CMD check to only Ubuntu on #355.

AARON-CLARK:
best attempt to solve merged with #359

PR #359, opened 2023-01-19
Increase shinytest2 timeouts 

borgmaan:
Increases shinytest2 timeouts across the board to address #354


AARON-CLARK (approved):
So hacky, but I like it.

PR #352, opened 2023-01-16
Rk 340 pkg del option

Robert-Krajcik:
Addresses Issue #340 Need admin package delete option

PR #365, opened 2023-01-23
Revert "Rk 340 pkg del option"

Jeff-Thompson12:
Reverts pharmaR/riskassessment#352

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: riskassessment
 Title: A web app designed to interface with the `riskmetric` package
-Version: 0.0.1.9024
+Version: 0.0.1.9023
 Authors@R: c( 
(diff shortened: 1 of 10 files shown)


AARON-CLARK (changes requested):
Hi @Robert-Krajcik, I was just starting a review your PR when @Jeff-Thompson12 finsihed his review and merged. No big deal, but I thought I'd add my comments here:

First, could we change the color of this button so that it really pops against the rest of the page? I think we should visually indicate that this button is "special". Maybe a light blue?

image

Secondly, I thought it probably matters to include the pkg's version number in this summary table since we are storing it in the DB:

image

AARON-CLARK:
eh, let's just close this revert PR and I'll open a new issue @Robert-Krajcik, sorry!

PR #367, opened 2023-01-24
tests for `auto_font()` function

AARON-CLARK:
Working towards more test coverage on #295


Jeff-Thompson12:
@AARON-CLARK The checks are pretty extensive for normal usage. Since we are exporting `auto_font()`, do we want to add any checks for invalid parameters?

AARON-CLARK:
> Since we are exporting `auto_font()`, do we want to add any checks for invalid parameters?

@Jeff-Thompson12, that would be a good idea. But do you think we need to export it? Maybe we don't. User's don't need it to run the app. Maybe if we start exporting the functions used to build cards, we should export `auto_font()`. Thoughts?

AARON-CLARK:
@Jeff-Thompson12 done!

AARON-CLARK:
This seems silly. You can only have @examples for an exported function? Oh well, I removed them with commit 64b4c1b

image

Jeff-Thompson12:
> This seems silly. You can only have @examples for an exported function?

@AARON-CLARK you can still have the examples if you use `@noRd` in the roxygen notes. That's what we have done with other functions.

AARON-CLARK:
Okay @Jeff-Thompson12, this is ready for a re-review. I'll leave it without examples for now.

AARON-CLARK:
Hey @Jeff-Thompson12, do you want to give this one final review & merge?

Jeff-Thompson12 (approved):
LGTM!

PR #374, opened 2023-01-27
Added new RValHub Slack badge

AARON-CLARK:
Let's folks know:
* this is a project sponsored/supported by the R Validation Hub
* there may be additional discussion on slack

The new `README` badge lineup looks like this:

image


borgmaan (approved):
Looking very pro!

PR #375, opened 2023-01-27
change border color for remove package button to background color

Robert-Krajcik:
Just this one small change!


Robert-Krajcik:
Now the ocular trauma is gone.
image

Issue #266, opened 2022-10-04
Export the pretty stuff


AARON-CLARK:
Basically, the `{riskassessment}` package could serve as a hub to extract all the sexy graphics we've developed into the app. That is, they can produce any cards & graphics programmatically for their own purposes, whatever they may be. Namely, I'm thinking of the community usage plot and also the any of the cards from any tab.

ScottSchumacker:
Hi @AARON-CLARK, Thank you! Yes, I spent some time learning about roxygen2 and roxygen comments and the different tags. It has been a while since I have worked on R packages but things are definitely more clear now. I have created a new branch "ss-266", created off of the dev branch that I will be working on for this issue.

AARON-CLARK:
closed with #377

PR #377, opened 2023-01-30
Issue 266: PR for ss-266New

ScottSchumacker:
* Added a new exported function called generate_comm_data() that builds the community usage data needed to produce the downloads by month plot (#266).
* Updated insert_community_metrics_to_db() internals to call generate_comm_data() (#266).
* Exported and updated the build_comm_plotly() function internals to call generate_comm_data as well as utilize a new parameter pkg_name (#266).
* Added tidyr as an import in the DESCRIPTION file
* Incremented the version in the DESCRIPTION file


AARON-CLARK (approved):
Looks great! Thanks @ScottSchumacker. I pushed a few more commits to make the `NEWS` file more succinct / brief and got rid of the `tidyr` dependency since `tibble()` is also exported through `dplyr` which is already on our `import`s list.

When ready, please follow up to this PR with a test for the `generate_comm_data()` and `build_comm_plotly()` functions.

AARON-CLARK (approved):

PR #392, opened 2023-02-06
Comm data can contain multiple versions per month, bloating downloads

### LINKED ISSUE #391: generate_comm_data left_join gives "matches multiple rows" warning
Some packages, like `Tplyr`, sometimes have multiple versions coming out on the same month.
The `versions_with_dates` DF, extracted from callling `get_latest_pkg_info(pkg_name)`, shows that for `Tplyr`:

image

I suggest using a different approach.
Instead of using `'left_join` we slot the version into the `pkgs_cum_metrics` data first 
and then summarize by month and year.

AARON-CLARK:
Closes #391 and also fixes a `testthat` test [warning](https://github.com/pharmaR/riskassessment/actions/runs/4057150136/jobs/6982536348#step:7:227):

image

Regarding community usage data, if there was more than one version release in the same month (for a lot of pkgs), so this PR implements a solution where the first and last version numbers from each month were concatenated together to summarize to one simple string. The community usage data really needs to be in the one row per year/month because that's the granularity that our download counts are summarized. We don't want to re-summarize by version, year, and month.

With the change, the new community usage plot looks like this:

image

Thanks for reviewing!


borgmaan:
@AARON-CLARK -- this looks good. Would you be up for adding a test to protect this from future regressions?

As we think about a pragmatic testing approach in a post- #295 world, one approach could be:

* Core feature of new functionality get tests
* Any bug fixes get an associated test

The latter can be a helpful way to establish a safety net as the application continues to expand in complexity.

AARON-CLARK:
> Would you be up for adding a test to protect this from future regressions?

I'm up for it.

> Core feature of new functionality get tests
> Any bug fixes get an associated test

Make sense to me.

borgmaan (approved):
💪 tests!

PR #394, opened 2023-02-06
Rename modules to follow `golem` naming convention

### LINKED ISSUE #384: Modules don't follow naming convention
Just noticed that our modules don't follow the convention to be pre-fixed with "mod_" which I think effects some golem functionality. A simple rename of these files shouldn't impact much, since the names of the actual server and ui functions won't be changing.

image

AARON-CLARK:
Fixes #384


borgmaan (approved):
👍

Issue #395, opened 2023-02-06
Loggit file is not being set


Jeff-Thompson12:
The function `golem::is_running()` does not return `TRUE` with how the app is constructed so the loggit file is not being set.

AARON-CLARK:
fixed by #396

PR #396, opened 2023-02-06
Set loggit file

Jeff-Thompson12:
Addresses #395


AARON-CLARK (approved):
@Jeff-Thompson12, confirmed working!

Issue #267, opened 2022-10-04
Decision Automation


AARON-CLARK:
Have user enter pre-set rules that automatically categorize packages into the "low", "medium", or "high" risk categories. We'd probably want to add an overall comment that mentions "the decision was selected automatically based on preset rules: {list the rule(s)}...

From a UI perspective, this could either exist in some new global location (like on the 'Assessment Criteria` tab) or in the Admin tab, but also on the upload tab for file-specific rules! This feature would justify keeping the upload tab around, but I suppose it could exist somewhere else.

AARON-CLARK:
It would also be nice to collect a consensus on the thresholds other companies are comfortable using to categorize pkgs into "low", "medium", and "high" risk labels. That means, it'd also be nice to grab a consensus on metric weights too, if one exist.

AARON-CLARK:
closed with #397

Issue #145, opened 2021-12-01
Risk re-calculation & Retaining pkg decisions


AARON-CLARK:
Users are going to strongly dislike re-evaluating potentially hundreds of previously reviewed packages every time the decision rules or metrics weights are tweaked and risk scores re-calculated, so I think we should consider adding options to retain final decisions for certain groups of packages. For example, the users could have some of these options: 

Drop the final decision for...

- No pkgs. That is, leave all final decisions
- Packages whose score would get worse (go up)
- Only the latest version of the package. IE Don't change a decision on an older version of the package we already reviewed. This is only applicable once we actually start leveraging pkg version.
- A list of user-defined packages. Perhaps we provide a multi-select where users can choose packages to drop decisions for?

AARON-CLARK:
@Jeff-Thompson12, these were the options I was mentioning in our meeting today. Looks like I wrote them down!

aclark02-arcus:
Now that #663 is integrated, I think this package retention modal should pop up when uploading a csv that has the `decision` column populated.

PR #397, opened 2023-02-06
Decision Automation

Jeff-Thompson12:
Implements #267


Jeff-Thompson12:
@AARON-CLARK This is still a work in progress (i.e. I still need to implement the decision automation in the reweights module and I have not written the test or documentation for added functions), but I would like you to take a look and see if this is matching the vision or #267.

AARON-CLARK (dismissed):
@Jeff-Thompson12, yes, yes it does! I really like how you implemented these ranges. Specifically, how each categories range is optional and how you can't overlap ranges. Plus, how you can leave gaps between ranges, indicating "grey" areas (see below)! Also love that you see the Decision made in the summary of uploaded pkgs & in the overall comments. Users will visibly see all these automatic changes and  great visible coverage that an automatic feature with this much weight really needs.

image

A few questions/ comments, some of which are unnecessarily in the weeds given this is still a draft, so sorry. But I thought I might as well mention them now!

* I would love to see the gear dropdown' panel a lot wider and the "Auto-Assign Risk Decisions For..." title much larger/bolder. Maybe an `h3()` or `h4()`?
* The gear will need to be an `admin` only feature. For `non-admin`s, it'd still be nice to see the Decision Automation panel, if applied. However, you'd have to lose the instructions to "click on the gear..." sentence. Do you think `non-admin`s should be able to turn the decision rules on and off?
* In the summary of uploaded packages, can we also include a count of "Decisions made" and a break down on if they were "Low", "Medium", and "High". If you want to get fancy, make these summary stats conditional on if they had any rules applied for them. Aka, if they didn't make a "medium" rule, don't show a summary stat for it. I think there is room for it directly to the right of the "status" summary stats.
* Eventually I think adding a "Decision Made Date" column will become important to the database overview. In addition WHO made the decision (which user) and even if it was an auto decision - the user value be "Auto Decision" or something.
* Just an idea, but do you think the user needs to confirm auto-decisions with some sort of confirmation modal that includes a sortable table with the package name, version, decision, and checkbox to accept or reject? Maybe something to consider in the future. I think it would be useful for an `ADMIN` trying to figure out what decision rules to choose.
* obviously, the number of categories should be flexible too in the future. For example, if the org just wants go/no-go categories. Or if they want 5 categories!

Jeff-Thompson12:
> * The gear will need to be an `admin` only feature. For `non-admin`s, it'd still be nice to see the Decision Automation panel, if applied. However, you'd have to lose the instructions to "click on the gear..." sentence. Do you think `non-admin`s should be able to turn the decision rules on and off?

This should already be true. Did I overlook something?

AARON-CLARK:
> This should already be true. Did I overlook something?

Sorry, I didn't test the `non-admin` experience!

Jeff-Thompson12:
> * I would love to see the gear dropdown' panel a lot wider and the "Auto-Assign Risk Decisions For..." title much larger/bolder. Maybe an `h3()` or `h4()`?

I originally had the box about as wide as the panel with the sliders side by side. Would that be preferable?

AARON-CLARK:
> I originally had the box about as wide as the panel with the sliders side by side. Would that be preferable?

hmmmm, I kind of like the idea of them being stacked, but I'm open to side-by-side. Go with what you feel looks best and I can provide my opinion if you need a tie-breaker.

Jeff-Thompson12:
@AARON-CLARK I made some additions not mentioned above.

- Added a reset button
- Added a confirmation modal for the decision rules

Below are my responses to your comments above

>I would love to see the gear dropdown' panel a lot wider and the "Auto-Assign Risk Decisions For..." title much larger/bolder. Maybe an `h3()` or `h4()`?

The label for the checkbox group has been made larger and bold. The window is now 600px wide.

>The gear will need to be an `admin` only feature. For `non-admin`s, it'd still be nice to see the Decision Automation panel, if applied. However, you'd have to lose the instructions to "click on the gear..." sentence. Do you think `non-admin`s should be able to turn the decision rules on and off?

Already implemented. Non-admin users see nothing if no rules have been added. If decision rules are present, a non-admin user will see the summary table of the rules.

>In the summary of uploaded packages, can we also include a count of "Decisions made" and a break down on if they were "Low", "Medium", and "High". If you want to get fancy, make these summary stats conditional on if they had any rules applied for them. Aka, if they didn't make a "medium" rule, don't show a summary stat for it. I think there is room for it directly to the right of the "status" summary stats.

The summary has been added. It will only show up if decision rules have been enabled.

>Eventually I think adding a "Decision Made Date" column will become important to the database overview. In addition WHO made the decision (which user) and even if it was an auto decision - the user value be "Auto Decision" or something.

I agree. Let's do that in another PR.

>Just an idea, but do you think the user needs to confirm auto-decisions with some sort of confirmation modal that includes a sortable table with the package name, version, decision, and checkbox to accept or reject? Maybe something to consider in the future. I think it would be useful for an `ADMIN` trying to figure out what decision rules to choose.

I think this functionality can be added later. We would only want the modal to pop up for admin users, and admin users can already reverse decisions. I kind of like that the loggit file records a decision being automatically made. Currently, we would log that an admin overrode the decision, which I think is preferred behavior.

With this in mind, we should probably add a not to the log file when rules are created/enabled or disabled.

>obviously, the number of categories should be flexible too in the future. For example, if the org just wants go/no-go categories. Or if they want 5 categories!

Yes, I tried to program with this somewhat in mind. We should probably think about implementing this functionality soon before it gets too hard to add.

AARON-CLARK (changes requested):
@Jeff-Thompson12, overall - amazing PR! I have a few suggestions below. Also, I wasn't able to review everything, so I'll come through with a f

What if I have packages already uploading in the database (perhaps I uploaded a CSV with no decision rules), and I re-load them CSV with decision rules activated/applied at a later date? Should we do nothing (which is the current functionality), or should those pkgs get decisions applied? I think if there existed no decision previously, I'd want those package decision rules applied. Or perhaps that's something we need to ask users in the dropdown panel? Basically, allow them to decide if we should apply decision rules on new pkgs only, all packages, or all packages with no previous decision.

image

Also, I tried to upload a duplicate pkg (`dplyr`) using the "type package here" text input, and it a strange output resulted. First, I noticed that the Decision column showed up, where it didn't show up when a CSV was uploaded. It also showed the decision made count  was NA, when technically the risk score is .13, so it should have fallen into the "Medium" group.

image

I uploaded one that didn't exist in my database, and I'm slightly confused by the Decisions made counter... I think I'm reading it right - there was 1 decision made, where 1 was high. But can we made create a little blank space/ margin between these stat counts? Maybe don't use the `-` either, just stick with the `:` so it should read `High: 1`. Also, I think it would give folks a little reassurance that everything worked correctly if they say the pkg score in the table next to the decision so that hey know the auto decision was applied appropriately. What do you think?

image

Can this dropdown panel get a darker border? It appears to bleed into the rest of the UI. Here's a before and after with me drawing on it. I like the clear separation.

image

image

Okay, I'll have more time to review tomorrow. Thanks for the progress!

Jeff-Thompson12:
> What if I have packages already uploading in the database (perhaps I uploaded a CSV with no decision rules), and I re-load them CSV with decision rules activated/applied at a later date? Should we do nothing (which is the current functionality), or should those pkgs get decisions applied? I think if there existed no decision previously, I'd want those package decision rules applied. Or perhaps that's something we need to ask users in the dropdown panel? Basically, allow them to decide if we should apply decision rules on new pkgs only, all packages, or all packages with no previous decision.

This makes more sense (to me) to tackle in #145.

AARON-CLARK (changes requested):
Hey @Jeff-Thompson12, here is my 2nd review. Just a few more comments.

1. I noticed that with these rules:

image

I did some re-weighting and noticed that packages with a score of 1.0 didn't receive a decision. In that respect, how is inclusion handled at the ends of each range. For example, right now, I have both the "low" and "medium" range on 0.20. So, if a pkg has exactly that score, will it be labeled with a "low" or "medium" decision. Should we default to inclusion on the top of the range? The "low" category could have inclusion on both ends, though a pkg score of zero is (probably) unattainable?

image

AARON-CLARK:
Hey @Jeff-Thompson12, just following up to you convo from today. This is what's still needed on this PR (and maybe a few more I didn't mention the first time):

### TODO:

- [ ] In the confirmation modal, lose the word "classification" in "Please confirm your chosen decision classification rules:". Also, change the footer text slightly: "Note: Once submitted, these rules will be applied to all new packages loaded into the app or when any metric re-weighting is performed."

- [ ] In the upload summary table, include the pkg score in a new column

- [ ]Also in the In the upload summary section, create more white space/ margin between stat counts summarizing the counts for each decision type. Right now, they are too close together, making it difficult to read quickly.

- [ ] Determine inclusion/exclusion of upper/lower bounds on decision range sliders. Is this something we let users edit/toggle?

### Future PRs:

- [ ] Add "Decision Made Date" and "Decision Made by" column to database view, including "auto-decision" as a possible user

- [ ] Address different routes users would want to overwrite decision in the database (#145)

- [ ] a new `run_app()` arg should be able to control the number of decision categories and category labels (using defaults "low", "medium", and "high")

AARON-CLARK (commented):
One quick question: why use `auto_decisions.json` to store these auto-decision ranges? Because it seems like a waste of a table to store it on the database? I'm thinking it might be better to have them in the database so that database downloads on the admin tab are complete and comprehensive. Thinking this will be best for admins/ devs.

Jeff-Thompson12:
> One quick question: why use `auto_decisions.json` to store these auto-decision ranges? Because it seems like a waste of a table to store it on the database? I'm thinking it might be better to have them in the database so that database downloads on the admin tab are complete and comprehensive. Thinking this will be best for admins/ devs.

I should have added my thoughts on this construction in the PR. I originally had it as an additional table but my thoughts were that decision automation rules could be larger than the project itself (i.e. shared across projects). By keeping it separate, it is treated more like meta data, which is how I personally view it. I also think that with its current construction, it is easier to manage and maintain.

AARON-CLARK:
> ...my thoughts were that decision automation rules could be larger than the project itself (i.e. shared across projects). By keeping it separate, it is treated more like meta data...

I feel like metric weights may also fall into this "metadata bucket" too. What do you think?

AARON-CLARK:
@Jeff-Thompson12, FYI there are some merge conflicts here. Also, if you update the `NEWS` and increment the version, this will be ready to merge in my opinion. 

Thanks!

Jeff-Thompson12:
> I feel like metric weights may also fall into this "metadata bucket" too. What do you think?

I did think about this and did feel like the weights were more like the decision rules than the package information. Currently the app is not set up to be pre-seeded with information. Maybe it is a both and. Maybe we should allow users to add meta data which will populate tables on the initial instance of the app, but all changes to those values be maintained in the database.

AARON-CLARK (approved):
👍🏼

Issue #399, opened 2023-02-07
renv/GHA Cleanup


borgmaan:
A few items to clean up in the `renv`/GHA realm:

- `fontawesome` should get upgraded to the compatible version (0.3.0) in [`renv.lock`](https://github.com/pharmaR/riskassessment/blob/dev/renv.lock#L752-L762) to avoid needing to re-install it during the execution ([here](https://github.com/pharmaR/riskassessment/blob/fb6b17a1df66ef6976321abe7a0505e2f21f98e4/.github/workflows/R-CMD-check-dev.yaml#L55), [here](https://github.com/pharmaR/riskassessment/blob/fb6b17a1df66ef6976321abe7a0505e2f21f98e4/.github/workflows/R-CMD-check-master.yaml#L54), and [here](https://github.com/pharmaR/riskassessment/blob/fb6b17a1df66ef6976321abe7a0505e2f21f98e4/.github/workflows/test-coverage.yaml#L32))
    - This will allow the app to be run without issue after `renv::restore()` and will save some time on each applicable GHA run 
- There is currently a [warning thrown](https://github.com/pharmaR/riskassessment/actions/runs/4115324874/jobs/7103884662#step:6:1) on all runs of `r-lib/actions/setup-renv`. This should be investigated and remedied if possible.

AARON-CLARK:
> * `fontawesome` should get upgraded to the compatible version (0.3.0) in [`renv.lock`](https://github.com/pharmaR/riskassessment/blob/dev/renv.lock#L752-L762) to avoid needing to re-install it during the execution ([here](https://github.com/pharmaR/riskassessment/blob/fb6b17a1df66ef6976321abe7a0505e2f21f98e4/.github/workflows/R-CMD-check-dev.yaml#L55), [here](https://github.com/pharmaR/riskassessment/blob/fb6b17a1df66ef6976321abe7a0505e2f21f98e4/.github/workflows/R-CMD-check-master.yaml#L54), and [here](https://github.com/pharmaR/riskassessment/blob/fb6b17a1df66ef6976321abe7a0505e2f21f98e4/.github/workflows/test-coverage.yaml#L32))

[...]

PR #398, opened 2023-02-06
Modify `renv` Rmd vignette

AARON-CLARK:
Provides more context to developers for onboarding purposes. Did I accurately describe a decent `renv` workflow when working amongst several developers?

PR #364, opened 2023-01-23
Add Administrative Tools and Options vignette

Robert-Krajcik:
Revising the older "managing userid and passwords" vignette.
Although I'm thinking about calling this "Administrative Tools and Options"
and covering both the "Credential Manager" and "Assessment Reweighting" tabs.

PR #400, opened 2023-02-07
Upgrade `{fontawesome}`

AARON-CLARK:
Fixes part of #399

This PR:

* updates `renv` lock file which is pointing at fa `v2.2.1` (really old) to `v0.5.0`
* removes custom GHA installation of fa `v0.3.0` in favor of the recently released `v0.5.0`
* Removes login note & console messages regarding installing `v0.3.0` and instead re-focuses on installing a recent version when it detects the problematic `v0.4.0` is installed
* adds fa to Suggests field
* Oh, it also updates our R version to `v4.2.2`

Note, once this if fixed, we can push through other PRs
* #364 
* #398

One problem I see with `v0.5.0` is that it produces a new warning in the console when a report is downloaded (which I'm not sure we have any control over):

image

Thoughts? We could go back to `v0.3.0` I suppose to avoid it?


AARON-CLARK:
Quick update, it appears the `xfun:isFalse()` issue stems from [this](https://github.com/yihui/xfun/pull/67) and [this](https://github.com/yihui/xfun/pull/66). Reverting back a release fixes the issue (for now).

borgmaan (approved):
👍

PR #408, opened 2023-02-13
Source `riskmetric` from GitHub

AARON-CLARK:
All developers (and users) should be using the latest GitHub version until the `riskmetric` team can push a new release to CRAN (which is coming soon). Note, the latest version on GitHub has the same version number as CRAN which is slightly confusing, so this PR just changes the source in the `renv.lock` file.

I also added a (hopefully temporary) note to the `README` to install `riskmetric` from GitHub.


Robert-Krajcik (approved):
Looks good to me!

Issue #412, opened 2023-02-13
Package metrics should match version of package in package table


borgmaan:
There is currently a bug that allows for metrics to be written to the `package_metrics` table to be derived from a version of the package that does not match the version logged in the `package` table. 

Details can be found in [this comment](https://github.com/pharmaR/riskassessment/issues/385#issuecomment-1428641592) on #385.

Robert-Krajcik:
Hi @borgman
Thanks for looking into this!

Borgman:
> Hi @Borgman Thanks for looking into this!

hey, he is @borgmaan :D I am Borgman

Robert-Krajcik:
Oops! Wrong Borgman.  Apologies @borgmaan

aclark02-arcus:
Closed with #413 and later with #385

Issue #385, opened 2023-02-02
package version discrepancy on upload


Robert-Krajcik:
Deleting package rvest.  Note version.
![image](https://user-images.githubusercontent.com/75215958/216400652-f13e5274-c4be-48a9-9910-9000b6dc09d2.png)
And then upload the package again:
![image](https://user-images.githubusercontent.com/75215958/216400719-a751bc6c-c549-4109-8e87-a951dd110bdf.png)
Note the version is one level back from the one just deleted.
Go to the sidebar and select the package, and the correct version is shown:
![image](https://user-images.githubusercontent.com/75215958/216400930-019b702c-9eb7-4b4a-9074-8f51ff5f001d.png)

borgmaan:
This is due to a bug in the way that `riskassessment` is handling versions and interfacing with `riskmetric`. I discovered this today when looking at #410.

Currently, `riskassessment` [inserts the latest CRAN version](https://github.com/pharmaR/riskassessment/blob/dev/R/utils_insert_db.R#L60-L66) as the package version in the DB. 

However, it [uses](https://github.com/pharmaR/riskassessment/blob/1cfb13e34ad4649fda1b63517792b8cd5b804ba3/R/mod_uploadPackage.R#L286) the version derived from a call to `riskmetric::pkg_ref` to populate the version shown in that summary table. 

Given the way `riskmetric::pkg_ref` works, this will default to a `pkg_install` source when the package is installed locally (as `rvest` is in our case with v1.0.2 being in our `renv.lock` file). You will see 1.0.2 show up in the summary table as that matches what is in your local library. However, it will be inserted to the DB as 1.0.3 and will show up as that version in subsequent activity. 

In addition to the bug you show here, this also leads to a severe bug related to storing `riskmetric` metrics and scores within the DB. `insert_riskmetric_to_db` also uses a ["bare" call to `pkg_ref`](https://github.com/pharmaR/riskassessment/blob/ef695c5942b72754e5b6f74107e3196d169bf77c/R/utils_insert_db.R#L147). In this case, **the metrics being written back to the DB can be for a different version of the package than what is logged in the `package` table in the database.** This is not the desired behavior. 

`shiny` is another good example where the repo's lockfile version is older than latest CRAN and `riskassessment` will log the latest CRAN version in the DB when inserting into the `packages` table but will insert metrics for the older, locally installed, version when populating the `package_metrics` table.

AARON-CLARK:
[...]

PR #413, opened 2023-02-14
Make app's pkg sources CRAN-focused (temporarily)

### LINKED ISSUE #410: Align `riskmetric` pkg repo source(s)
I noticed for packages recently updated on CRAN, all the `riskmetric` metrics were borked:

image

After a little bit of digging, I found that `riskmetric::pkg_ref()` is smart enough to look up our environment's repo source, which so happens to be pointing to a CRAN snapshot for `renv.lock` purposes:

image

As a result, the pkg info we scrape by hand from CRAN clashes with the info that `riskmetric` provides, and displays faulty metric & pkg scores. However, there is a way to point `riskmetric::pkg_ref()` at a different source using the `repos` arg:

image

And that fixes the problem for this scenario:

image

However, I'm not sure we can specify all the repo source contexts that exist... For example, Bioconductor, GitHub, etc

image

Any ideas?

AARON-CLARK:
Why? Well, to gain unity on  version discrepancies, fixing:
* #412
* #385

And also, to fix part of #410, since any pkg that has a newer release than what's in our `renv.lock` file get's borked metrics on upload. More discussion is yet to be had on how to fix #410 for the future.

### DIFF
--- a/R/mod_uploadPackage.R
+++ b/R/mod_uploadPackage.R
@@ -251,7 +251,7 @@ uploadPackageServer <- function(id, user) {
             
             if (grepl("^[[:alpha:]][[:alnum:].]*[[:alnum:]]$", uploaded_packages$package[i])) {
               # run pkg_ref() to get pkg version and source info
-              ref <- riskmetric::pkg_ref(uploaded_packages$package[i])
+              ref <- riskmetric::pkg_ref(uploaded_packages$package[i], repos = c("https://cran.rstudio.com"))
             } else {
               ref <- list(name = uploaded_packages$package[i],
                           source = "name_bad")
--- a/R/utils_insert_db.R
+++ b/R/utils_insert_db.R
@@ -144,7 +144,7 @@ insert_riskmetric_to_db <- function(pkg_name,
     db_name = golem::get_golem_options('assessment_db_name')){
 
   riskmetric_assess <-
-    riskmetric::pkg_ref(pkg_name) %>%
+    riskmetric::pkg_ref(pkg_name, repos = c("https://cran.rstudio.com")) %>%
     dplyr::as_tibble() %>%
     riskmetric::pkg_assess()
   
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: riskassessment
 Title: A web app designed to interface with the `riskmetric` package
-Version: 0.0.1.9032
+Version: 0.0.1.9033
 Authors@R: c( 
     person("Aaron", "Clark", role = c("aut", "cre"), email = "aaron.clark@biogen.com"),
     person("Robert", "Krajcik", role = "aut", email = "robert.krajcik@biogen.com"),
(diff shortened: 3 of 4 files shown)


AARON-CLARK:
@Jeff-Thompson12 & @borgmaan, how does this look for the temporary fix @borgmaan proposed in #410. Note: using `riskmetric::pkg_ref(..., source = "pkg_cran_remote")` didn't work, so I opted to use `riskmetric::pkg_ref(..., repos = c("https://cran.rstudio.com"))` instead.

Also, the R CMD Check failure seems to be they typical `shinytest2` timeout failure. I keep re-running, but the ✔️ is elusive.

R/mod_uploadPackage.R
```diff
@@ -251,7 +251,7 @@ uploadPackageServer <- function(id, user) {
             
             if (grepl("^[[:alpha:]][[:alnum:].]*[[:alnum:]]$", uploaded_packages$package[i])) {
               # run pkg_ref() to get pkg version and source info
-              ref <- riskmetric::pkg_ref(uploaded_packages$package[i])
+              ref <- riskmetric::pkg_ref(uploaded_packages$package[i], repos = c("https://cran.rstudio.com"))
```
borgmaan, line 254: I would suggest:

```r
ref <- riskmetric::pkg_ref(uploaded_packages$package[i], source = "pkg_cran_remote", repos = c("https://cran.rstudio.com"))
```

So you can also patch #412 and #385

R/utils_insert_db.R
```diff
@@ -144,7 +144,7 @@ insert_riskmetric_to_db <- function(pkg_name,
     db_name = golem::get_golem_options('assessment_db_name')){
 
   riskmetric_assess <-
-    riskmetric::pkg_ref(pkg_name) %>%
+    riskmetric::pkg_ref(pkg_name, repos = c("https://cran.rstudio.com")) %>%
```
borgmaan, line 147: Same thing here:

```r
ref <- riskmetric::pkg_ref(pkg_name, source = "pkg_cran_remote", repos = c("https://cran.rstudio.com"))
```

borgmaan (changes requested):

R/utils_insert_db.R
AARON-CLARK: resolved with [32e8074](https://github.com/pharmaR/riskassessment/pull/413/commits/32e807439ca13f553d942d1e729014c4d4a1bdaf)

R/mod_uploadPackage.R
AARON-CLARK: Resolved with [32e8074](https://github.com/pharmaR/riskassessment/pull/413/commits/32e807439ca13f553d942d1e729014c4d4a1bdaf)

AARON-CLARK:
@borgmaan, request changes on the 'source' arg are summitted. Though I'm not sure why the GHA's aren't kicking off on that commit (32e8074). I pushed another dummy commit to re-launch them, and that didn't appear to work either.

borgmaan (approved):

Issue #415, opened 2023-02-15
Report Bugs is no longer URL


Jeff-Thompson12:
The value for `has_bug_reports_url` being returned from `riskmetric::pkg_assess()` is no longer a URL, but a 0-1 values indicating whether a bug reports url is present in the repository.

Robert-Krajcik:
Closed with PR #416 Fix Report Bugs

PR #414, opened 2023-02-14
Prep for `v0.1.0` release

AARON-CLARK:
This PR:
* Accomplishes part of #402
* Accomplishes the html part of #378 
* Closes #379 
* Adapts reports to new has_report_bugs_url assessment value
* makes sure all .Rd files are up to date, making sure they have all the correct tags, (for example `@return` tags for exported functions)
* removed notes from R CMD Check - specifically addressed new global variables
* increments pkg to `v0.1.0`
* streamlines the `NEWS` file for better user consumption
* updates the `README` with best practice installation instructions (aka, cloning the repo)
* Fix any `TODO` items (in scope for this release)
* removes all `pre_auth_user` arg "stuff"
* checks URLs
* Fixes spelling
* re-organizes authors by "time on project", role/ status, and entity (human vs org - moving orgs to the bottom)
* moves `pkgload` to suggests as it's only needed in the `app.R` file.

PR #416, opened 2023-02-15
Fix Report Bugs

Jeff-Thompson12:
Addresses #415


Jeff-Thompson12:
This update does require the assessment database to be reinitialized since `has_bug_reports_url` was coded as a URL in the SQL files. Although I suppose one could run code to update that table without deleting `database.sqlite`.

I checked the value of `has_news` and verified that the field is supposed to be the "number of discovered NEWS files" so no change is needed.

Robert-Krajcik (approved):
Changes are working.  Getting TRUE or FALSE, depending on package.
image

R/mod_metricGrid.R
```diff
@@ -50,7 +50,9 @@ metricGridServer <- function(id, metrics) {
         metricBoxServer(id = metric['name'],
                         title = metric['title'],
                         desc = metric['desc'],
-                        value = metric['value'],
+                        value = dplyr::case_when(metric['name'] != 'has_bug_reports_url' ~ metric['value'],
+                                                 metric['value'] == "1" ~ 'TRUE',
+                                                 TRUE ~ 'FALSE'),
```
AARON-CLARK, line 55: FYI @Jeff-Thompson12, this needs to be implemented in each report as well, since the change was implemented at the metricBox-level (which is appropriate) and each report Rmd file defines a function on how to build the boxes. I'm adding this to #414

Jeff-Thompson12: Okay, I will also add a safety valve in for the URL argument being passed as well if not already done.

PR #429, opened 2023-02-20
Remove `pre_auth_user` from list of arguments in `run_app()`

Jeff-Thompson12:


Jeff-Thompson12:
@Robert-Krajcik I checked that the option is false in the event that a user does not use the app.R file provided and the option is not set.

Robert-Krajcik (approved):
LGTM!

AARON-CLARK:
@Jeff-Thompson12, so cool. `pre_auth_user` is like a secret hidden arg.

PR #430, opened 2023-02-22
Utilize temporary local libs to enable pkg_ref(..., source ="pkg_install")

borgmaan:


borgmaan:
This is working if pulled down and run locally. The tests will fail as they are installing some heavy packages (e.g. `tidyr`) and the time it takes to install these packages (to enable `source = "pkg_install"` exceeds the `shinytest2` timeouts. 

The intent here is to show an approach of using a temporary local library where packages are installed and used as the source for calls to `riskmetric::pkg_ref`. Getting this "right" from a UX perspective may be challenging as the package install is quite slow. However, I see something like this as a path forward to solving things like #52, #412, [this discussion](https://github.com/pharmaR/riskassessment/issues/410#issuecomment-1428663083), etc.

Jeff-Thompson12:
@borgmaan were you wanting to expand upon this draft PR? Or are you happy for someone else to take it over?

I did see that you were allowing the user to point to a library. I think I got confused with comments earlier because this PR will always install the added packages.

borgmaan:
> @borgmaan were you wanting to expand upon this draft PR? Or are you happy for someone else to take it over?
> 
> I did see that you were allowing the user to point to a library. I think I got confused with comments earlier because this PR will always install the added packages.

@Jeff-Thompson12 -- I  don't have bandwidth to actively develop this POC at the moment so all good for someone to take over. Happy to help brainstorm approaches/workflows (e.g. pointing at local libs, allowing for "cohorts" of versions of packages assessed together, etc) either on GH or during a call.

Jeff-Thompson12:
@AARON-CLARK, cool! I guess this means that we should be prioritizing installation over other sources since these features rely on that to work.

Jeff-Thompson12:
Closing because we are no longer going this direction.

PR #437, opened 2023-02-22
Fix 'Package downloads' bug

AARON-CLARK:
I just realized that there has been a major bug in whats reported in the card for "the number of downloads in the last 12 months" card on the Community Usage tab. Instead of summarizing the last 12 months, we just summarized all months from the last year. So if you run the code today, Feb 2023, it would just sum all of 2022, ignoring Jan 2023. Whoops!

For `stringr` package, before the fix:
image

After fix:
image


Robert-Krajcik (approved):
LGTM!

PR #439, opened 2023-02-23
Rk 417 pkg names

### LINKED ISSUE #417: Don't alter case on pkg names
We should avoid messing with the capitalization of package names, since they are case sensitive. Note: I'm pretty sure this is happening by default thanks to our `{bslib}` theme selection. The below snapshot is from an html report, but altered case may exist in other areas of the app.

image

Robert-Krajcik:
closes #417


AARON-CLARK (changes requested):
Hi @Robert-Krajcik, this looks really good! Wondering if you could also make this text "asis" as well? And maybe add the curly braces onto `riskmetric`? 

image

I tried all three report types, and it looks like HTML is the only one effected. Here's the Word and PDF output:

image

Robert-Krajcik:
Hi @AARON-CLARK 
Yes I can take a look at this in a short while.
I think these uppercase are coming from `mod_downloadhandler.R`

Robert-Krajcik:
Ready for another review, @AARON-CLARK

AARON-CLARK (approved):
Looks good!
image

PR #440, opened 2023-02-23
Documentation improvements

### LINKED ISSUE #411: README: Poll users how they use the app
For companies who used {riskmetric} as part of their validation effort/process, we need to collect data to identify a consensus on:
1.	Metric weights used
2.	Pkg score thresholds to make go/ no-go decisions

On #1, I know I’ve seen a few presentations where companies (like Merck & Roche) shared the metric weights they use, but has any of that information been catalogued somewhere? 

#2 is important because the primary goal of the {riskassessment} app is for companies to categorize each package into “low”, “medium”, or “high” risk. More than likely, we assume orgs rely heavily on the pkg score produced by {riskmetric} to make this decision. For example, perhaps if the pkg score < 0.25, then it’s considered low risk and thus, doesn’t get filtered out of their validation process just yet. Conversely, maybe if the pkg score is > 0.50, there is no way they’d include such a risky package in their validated environments. And thus between 0.25 and 0.50 is the grey zone where additional research & evaluation is needed.

Any other additional questions we want to know about our users? We can have a survey link on the README, but we'll also reach out to RValidationHub Use-Case companies.

AARON-CLARK:
PR accomplishes:
* adds the user feedback survey to the `README` (Closes #411)
* updates 'Get Started' vignette to include blurb on Decision Automation
* updates 'Admin Tools and Options' vignette to include larger blurb on Decision Automation
* updates 'Summary of Uploaded Package(s)' section to allow for more rows per page in DT object


Robert-Krajcik (changes requested):
Just a couple of suggestionss
Can you add "decision automation" to the Introduction part of "Administrative Tools and Options"?
image

And while you are here, we should make these "remove button" comments consistent
Do you prefer "red remove button" and "dark blue change password" button,
or just simply "remove selected users" and "change password"?


image

AARON-CLARK:
@Robert-Krajcik, thanks for the review! I've updated the vignettes accordingly.

AARON-CLARK:
@Robert-Krajcik, okay ready for a 3rd review! I colored lots of text, and made the text all caps to match the button.

Robert-Krajcik (approved):
Oh Yeah This looks really great now.
Thanks for making these changes @AARON-CLARK

Issue #424, opened 2023-02-20
CRAN logs website down


Jeff-Thompson12:
To populate the community usage metrics in the application, we are running the function `cranlogs::cran_downloads()`. It pulls the download information from [here](http://cranlogs.r-pkg.org/downloads/daily/). I personally have experienced this webpage being down several times in the past week. This results in no data being pulled in.

I think it would be nice to alert the user (or completely disable package upload) if needed websites are returning 404 errors.

AARON-CLARK:
@Jeff-Thompson12, I had this happen to me as well. Pretty frustrating because the progress bar took forever.

Forever, we've actually been kind of naughty by deviating from `riskmetric` and [summing up the number of downloads](https://github.com/pharmaR/riskassessment/blob/d31168373ae98c321f1f3e100c317863bf120cbc/R/utils.R#L328-L336) in the last 12 months ourselves for the Community Usage tab, ignoring the what `riskmetric` inserts into the DB. We do this so that our month-by-month plot doesn't include partial months, and we wanted the sum of the last 12 months displayed on the plot so our numbers checked out with what's displayed in the card. So, if we can't pull from `cran_downloads` due to a 404 error, maybe we don't stop the upload to db and just display the `riskmetric` assessment value, which should exist in the DB (as long as `riskmetric` isn't also getting denied with a 404 error from whatever source they are using). Just note, the risk score is always based on `riskmetric`s assessment and not ours.

FYI, as a result of inspecting this, I also just realized that there has been a major bug in what we report on the card for the number of downloads in the last 12 months. Instead of summarizing the last 12 months, we just summarized all months from the last year. So if you run the code today, Feb 2023, it would just sum all of 2022, ignoring Jan 2023. Whoops! The problem would get worse the farther you are into the year. For example, if you ran an assessment in Dec 2022, it would only sum months from 2021! Yikes. PR #437 opened to fix. Note this doesn't fix this issue though.

AARON-CLARK:
closed with #446

PR #446, opened 2023-02-28
Check URLs used to populate metrics

Jeff-Thompson12:
Addresses #424

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: riskassessment
 Title: A web app designed to interface with the `riskmetric` package
-Version: 0.1.0.9007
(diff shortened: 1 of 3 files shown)


Jeff-Thompson12:
I have added in the ability for the user to choose to be alerted when the unreachable URL can be reached (pinging approximately once a minute). I do not know if this really adds value since I don't know how long we anticipate user sessions lasting but will leave that up to the reviewer(s).

Robert-Krajcik (approved):
I disconnected from the internet and tried loading a package.
Then I received this modal.
image
Then I reconnected.  Of course, I had to "tickle" the `riskassessment` app (type in the package name again) and I got this other modal
image

Looks like its working!

PR #448, opened 2023-03-07
Add button that redirects to pkg metrics

Robert-Krajcik:


Robert-Krajcik:
What I have right now basically works.  I just need to figure out how to get back to databaseView from maintenanceMetrics

AARON-CLARK (changes requested):
Yessss, I think it works really well and is very quick & "snappy". It's a great addition!

I have 3 requests:
1. Could you change the Column name from "Explore Metric" to "Explore Metrics" plural?

2. It's hard to tell the button icon is a microscope, so perhaps just an arrow pointing to the right would be more appropriate?


3. I don't think we need a button that redirects users back to the database view from the Maint Metrics tab. If they want to view the Database, they should just click on the database tab. I think adding another button here is redundant and just makes the page unnecessarily busy.

Robert-Krajcik:
All set for another review, @AARON-CLARK

AARON-CLARK (approved):
👍🏼

PR #449, opened 2023-03-13
upgraded the RVersion to 4.2.2 in the renv.lock file

Xyarz:
Increased the R Version on the renv.lock file to 4.2.2


AARON-CLARK (approved):
Decided to include a line on the `news` file for this upgrade. Thanks! Looks good!

PR #453, opened 2023-03-14
Database Overview: Initial export

### LINKED ISSUE #383: Include Database Summary stats and export options
As we think about the final product that `riskassessment` app produces, we should also remember that validation efforts are a complicated and labor-intensive process that involves many different personnel to get involved. Thus, the end product can't solely be the detailed reports at package-level. To appease all parties (IT, management, others), we'll also need a high level summary of everything in the database, so I'd propose the database receives a few improvements and can be exported as a report itself.

### Table edits: 
- [x] New column: "Decision date" (this may need to get added to the db... not sure)
- [ ] Allow filtering (which is a built-in feature) in DT. We just have it turned "off" right now. If there are 500 packages in here, or all of CRAN, users may want to them up quickly

### New cards or plots:
- [ ] Total Count of pkgs in DB
- [ ] Count (and %s) of pkgs by decision made / not made
- [ ] Count (and %s) of pkgs by decision
- [ ] Histogram showing the distribution of pkg scores by decision category, overlain with lite opacity so we can see each group. It would be nice to have some interactivity but that's not a requirement.
- [ ] visual of decision

### Export features:
- [ ] First step, easy `DT` based Excel and CSV download button
But in the future, we users should be able to download a report that includes the DT object plus cards & Viz objects
- [ ] Users should be able to toggle some checkbox that specify if they just want the list of pkgs or visuals, or both
    - Ex checkbox: "Include visual stats & plots"
- [ ]  The entire report could be filtered by Score, Decision Made, or Decision, so we'd need a widget to include pkgs...
    - with scores between x_low and x_high. It could be a slider with two tickers set to defaults 0 and 1. But the user may move it to 0 and .25 for example.
    - for those with Decisions made (or not made). Would want a radio button here.
    - for those with a certain decision. For example, a groupCheckboxInput() with just "low", "medium", or "high"
- [ ] In addition to html, docx, pdf export formats, we should also have Excel as an option so folks can just get a long list of pkgs. I assume this will be the most popular option. If excel is chosen however, the checkbox for "include visual stats & plots" should be unchecked and disabled.
- [ ] The output file should contain the date and all the other report metadata we include in the package-level report, in addition to any filters applied on the database tab for the new widgets mentioned above.

AARON-CLARK:
Partially closes #383 by including:
* really low hanging fruit of adding an excel/csv download
* cleans up some header inconsistencies from tab to tab
* made 'Explore Metrics' button more compact so that more pkgs can fit on one page, reducing user scroll
* adds search capability so that users can find a package faster

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: riskassessment
 Title: A web app designed to interface with the `riskmetric` package
-Version: 0.1.0.9010
+Version: 0.1.0.9011
 Authors@R: c( 
     person("Aaron", "Clark", role = c("aut", "cre"), email = "aaron.clark@biogen.com"),
     person("Robert", "Krajcik", role = "aut", email = "robert.krajcik@biogen.com"),
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: riskassessment
 Title: A web app designed to interface with the `riskmetric` package
-Version: 0.1.0.9009
+Version: 0.1.0.9010
 Authors@R: c( 
     person("Aaron", "Clark", role = c("aut", "cre"), email = "aaron.clark@biogen.com"),
     person("Robert", "Krajcik", role = "aut", email = "robert.krajcik@biogen.com"),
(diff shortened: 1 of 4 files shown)


Robert-Krajcik (approved):
Looks Good to me!

PR #455, opened 2023-03-14
fix failing test, prefix appDriver with 'shinytest2::'

Robert-Krajcik:


Jeff-Thompson12:
@Robert-Krajcik I thought changing the base would be enough to trigger the master R-CMD check, but apparently that is not the case. Can you try pushing a small commit? Maybe some white space edits?

borgmaan (commented):
👍

borgmaan (approved):

PR #462, opened 2023-03-20
fix chromote error view() shows websockets disconnection

Robert-Krajcik:
`remotes::install_github("rstudio/chromote@origin-testfix")`
results in updating `chromote` 0.1.1 to 0.1.1
Need to update renv.lock

This now works.  Try it!
```
cs <- chromote::ChromoteSession$new()
cs$view()
```


Robert-Krajcik:
There are other `testthat` errors that need to be addressed in another PR!

PR #473, opened 2023-03-27
Fix bug on boundary decision categories

Jeff-Thompson12:


Jeff-Thompson12:
I would kindly request the reviewers to ignore the failed test on this PR because this PR did not introduce them and is a quick patch for a broken feature that needs repair.

Jeff-Thompson12:
If you try to add a decision rule for "Low Risk" or "High Risk" with the default configuration, it will break the app right now because the input slider is only returning one value. I have added back in the boundaries of 0 and 1 respectively.

Robert-Krajcik:
HI @Jeff-Thompson12 
PR #460 add decsion cols got merged today.
Did you try deleting database.sqlite?

Jeff-Thompson12:
Yes, but that would be unrelated to this PR and these failing tests because the bug is completely contained within the decision automation module. This bug got introduced with #459.

If you look at the files changed, you can see that only one line is being changed.

Jeff-Thompson12:
@AARON-CLARK & @Robert-Krajcik I suppose if you want me to fix the failing tests introduced with #460 in this PR, I can. Just let me know because this patch needs to be committed before our meeting tomorrow.

Robert-Krajcik (approved):
Tested, and it works.

PR #474, opened 2023-03-27
Fill gaps between decision rules instead of using uniform distribution

Jeff-Thompson12:


Jeff-Thompson12:
Yeah, you cursed my brain over the weekend and I knew the only way I was going to be allowed to stop thinking about it was to do it.

PR #488, opened 2023-04-12
Update db-config.yml

Jeff-Thompson12:
This PR changes the configuration of decision categories and rules to be a list instead of separate values. I believe this will aid in adding more complexity down the road (e.g. metric weights, colors, etc.)

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: riskassessment
 Title: A web app designed to interface with the `riskmetric` package
-Version: 0.1.1.9004
+Version: 0.1.1.9005
 Authors@R: c( 
     person("Aaron", "Clark", role = c("aut", "cre"), email = "aaron.clark@biogen.com"),
     person("Robert", "Krajcik", role = "aut", email = "robert.krajcik@biogen.com"),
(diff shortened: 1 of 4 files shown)


Jeff-Thompson12:
Hmm...it looks like you might have highlighted an unintended issue without realizing it. I think the reason it is not working is because of the indents. `rules` is a subcategory of `decisions`.

Jeff-Thompson12:
@AARON-CLARK Do you think this is an issue that can be corrected with documentation? I could also send a message to the console. I think we want to move away from people initializing the application with `riskassessment::run_app()` anyways. Thankfully we have the initialization already written as a function.

AARON-CLARK:
> Do you think this is an issue that can be corrected with documentation? I could also send a message to the console.

Yes, both! Some sort of warning (or error) message should probably be thrown.

Jeff-Thompson12:
I think the documentation should wait until we finish adding the other items to `db-config.yml` from our roadmap.

I have added some simple checks to let the user know that unexpected behavior may have occurred.

AARON-CLARK (changes requested):
This config file didn't produce any messaging in the console:
```yml
default:
  assessment_db: database.sqlite
  credential_db: credentials.sqlite
  decisions:
    categories:
      - Low Risk
      - Medium Risk
      - High Risk
  rules:
    Low Risk:
      - 0
      - .1
    High Risk:
      - .7
      - 1
```

image

AARON-CLARK (approved):
Just kidding! Works like a charm!

Issue #427, opened 2023-02-20
Admin tools: missing icons


AARON-CLARK:
![image](https://user-images.githubusercontent.com/17878953/220141583-45e9e7e6-3fef-491d-a091-6cbd8356b43a.png)

![image](https://user-images.githubusercontent.com/17878953/220141793-56343299-7e7d-449c-92a5-16612b360ea4.png)

Jeff-Thompson12:
So these icons are deprecated. If you install `fontawesome` version `0.3.0` (and fix a few bugs...), these icons show up again. Of course we have other icon issues elsewhere in the app when we do that.

Probably the most fun aspect of this is that these icons are a part of `shinymanager`, which means the actual fix would need to come from them. We already have a number of lines of code designed to "fix" issues we encountered by using a more recent version of bootstrap.

AARON-CLARK:
closed with #502

PR #496, opened 2023-04-13
Add fontawesome v4.7

Jeff-Thompson12:
Addresses #427


Jeff-Thompson12:
I am having difficulty suppressing the messages to the console:
```
The `name` provided ('pencil-square-o') does not correspond to a known icon
The `name` provided ('trash-o') does not correspond to a known icon
```
If any one knows how to handle this, I would appreciate some input.

Jeff-Thompson12:
An alternative approach could be to downgrade `fontawesome`, but I think we would want to avoid that solution.

AARON-CLARK (approved):
It works

Issue #463, opened 2023-03-20
Need to allow for custom location and name for `loggit.json`


AARON-CLARK:
Similar to `database.sqlite`, users can rename the assessment db to whatever they like using the `assessment_db_name` argument in `run_app()`. We need a similar functionality for `loggit.json`... mostly because the user should have complete control over where this is stored. In other words, they'll need to run something like `run_app(log_name = "./dev/mylog.json"`.

PR #497, opened 2023-04-13
Add log file to configuration

Jeff-Thompson12:
Addresses #463


AARON-CLARK:
@Jeff-Thompson12, looks like there is a "legit" failed test here.

@ScottSchumacker, do you have time to look at this one today?

Jeff-Thompson12:
> looks like there is a "legit" failed test here.

It's the second time I've seen this test fail in this spot (Note that it passed no issue in test-coverage). I think it's because we are dynamically inserting UI if the user is admin, so if they virtual machine is running a little slow, it might fail. I've added in a wait. We will see if that resolves the issue.

AARON-CLARK (approved):
👍🏼 👍🏼

PR #501, opened 2023-04-14
Fix tests failing on macOS and windowsOS

Jeff-Thompson12:
There seems to be tests that perform differently on macOS, windowsOS, and ubuntu. This PR is trying to resolve those issues.


Jeff-Thompson12:
It appears that the main culprit in the tests failing on the Windows OS is the initialization of the application (i.e. `initialize_raa()`). This PR has put some measures in place to reduce that burden, but we should consider letting a user initialize their configured environment outside of starting up the application.

borgmaan:
> I think we have finally achieved passing tests. Ending solution is a bit hokey, but I personally think it is okay.

*pragmatic ;)

Jeff-Thompson12:
Not really sure why `macOS` test started failing on `test-uploadPackage`. I am done with this PR for the time being though. I will keep an eye on it to see if it persists.

PR #502, opened 2023-04-14
Update `{shinymanager}`

Jeff-Thompson12:
Better solution to #427 

This is a better solution than adding all of fontawesome 4.7.1 to our application. Turns out the good folks over at `shinymanager` updated their package with the newer convention for icons from fontawesome.


AARON-CLARK (approved):
👍🏼 🥇

Issue #465, opened 2023-03-21
Custom Decision Color-coding


Jeff-Thompson12:
Allow the user to select their own colors applied to decision categories.

Only admin level users should be able to change/set colors. The color chosen should be stored to be used across sessions. This is meant to be an extension of #459, so it should also allow the user to set colors in the configuration of the application.

AARON-CLARK:
closed with #522

PR #510, opened 2023-04-26
Custom Decision Colors

Jeff-Thompson12:
Addresses #465

### DIFF
--- a/R/mod_decision_automation.R
+++ b/R/mod_decision_automation.R
@@ -307,7 +307,7 @@ mod_decision_automation_server <- function(id, user){
             right = TRUE,
             width = '600px',
             inputId  = ns("auto_dropdown_2"),
-            tooltip = shinyWidgets::tooltipOptions(title = "Click here to add/adjust decision automation rules.", placement = "left")
+            tooltip = shinyWidgets::tooltipOptions(title = "Click here to edit the decision category table.", placement = "left")
           )
         )
       })


Jeff-Thompson12:
This PR has an important change to the `package` table in the assessment database. Since we are tracking the decision categories inside of it's own table, the `package` table is no longer storing decisions. Instead it is storing a foreign key to the `decision_categories` table.

Jeff-Thompson12:
I chose the programming direction of expanding the database view tab in the application. I anticipate as we continue developing and allowing the user to have more control over the database that this section will continue to grow.

This also caused me to change how the decision automation module is being called. Since I needed the UI element in the upload packages tab to communicate with the UI in the database view, I am now calling the module once in the server.

Jeff-Thompson12:
I will wait to update the NEWS and version number until after the PR has been reviewed.

Robert-Krajcik (approved):
Looks good to me, Jeff!

AARON-CLARK:
Hi @Robert-Krajcik. Just a reminder, since this PR solves #465 which is labelled as an "enhancement", I'm supposed to give it a review before merging too! @Jeff-Thompson12, just a couple brief comments/ questions on this PR but first let me say: wow, it's amazing how many changes are required in the repo to institute this new feature. It touches a lot of code. Great work & thanks for the PR! It's a great addition.

## 1: Edit color UI location

> I chose the programming direction of expanding the database view tab in the application. I anticipate as we continue developing and allowing the user to have more control over the database that this section will continue to grow.

I'd argue it'd make more sense to create a new tab under the `Admin tools & Options` tab since it's supposed to be an `admin` or `lead` privilege, no? I'd argue anything that shows up in the config file probably deserves a spot there to have all those handy tools in one location of the app.

image

Also, I'm starting to get annoyed by this `shinyWidgets::dropdownButton()`. Perhaps we could transition to a collapse/ expand style ui? Certainly an out-of-scope request - could be addressed in another PR.

Another random comment: the "Check out the app's code" banner should probably shifted to the bottom of the page, just like on every other tab.



## 2. No color picker on upload pkgs tab
Was the color picker UI intentionally left off the upload pkgs tab? If we are allowing admins to update decision automation here, we could/should probably be consistent and allow them to update the colors too.

image



## 3. Adjusting the risk score's color gradient
I'm torn on whether we should be making the pkg score colors on the database view adjust to the new decision category colors chosen. In the example below, I made all the colors slightly more pale, but the risk score colors are still really bold:

image

In reality though, people may not keep with the same old green, yellow, red color scheme. They may use pink, purple, grey - who knows! Here's an example of switching the "high risk" category to grey. The risk score color really diverges from the decision category which won't make much sense visually.

image

I know we are using this code to set the color palette, so in theory, we could just use `colorRampPalette()` on whatever colors are chosen for the decision categories.

```r
# colorspace::darken(viridisLite::turbo(11, begin = 0.4, end = .8225), .25)
setColorPalette <- colorRampPalette(c("#06B756FF","#2FBC06FF","#67BA04FF","#81B50AFF",
"#96AB0AFF","#A99D04FF","#B78D07FF","#BE7900FF","#BE6200FF","#B24F22FF","#A63E24FF"))
```

The controversy here is what if the config file has these decision categories?

```yml
decisions:
    categories:
      - GxP Compliant
      - Needs Review
      - Not GxP Compliant
```
Does it make sense to ramp the colors between these three categories? Namely, we broke the likert scale with the "Needs Review" category, so ramping the colors between all three would say that the "needs review" color is halfway between "compliant" and "non-compliant" isn't necessarily true. Maybe a simple solution could be a `shinyWidgets::materialSwitch()` that simply asks the user if they'd like to "Render color gradient for risk scores using custom decision category colors", shifting the decision to the user. In most cases, I think they'd oblige, so it could default to the "on" position. What are your thoughts?




## 4. Applying risk & decision colors on pkg upload tab

I know it's probably a pain, but for completeness, we should probably start performing/sharing the same coloring actions on the upload pkgs tab too: 

image

Jeff-Thompson12:
@AARON-CLARK My bad. I didn't add you to the Reviewers list. I thought about it while the PR was in draft but forgot when I marked it ready-for-review.

1. Edit color UI location
> My thoughts were that as we started adding more elements to the database, it would be natural to find the editing options there as well. This makes sense to me with moving away from the admin/non-admin setup, where admin tools would be related more to management of users instead of the database.

> I was messing with the github link at the bottom of the page before. I don't like that it's the bottom of the HTML page and not the bottom of the visible page. I was going to tinker with it in another PR. I do agree that the dropdown covering it is undesirable behavior.

2. No color picker on upload pkgs tab
> I left it off here because the UI element here was focused on the decision automation rules. I felt that adding more would clutter the view on that tab. At least that's my reasoning.

3. Adjusting the risk score's color gradient.
> I contemplated this but thought it would make more sense to separate it and attach it as a feature related to the risk metrics themselves instead of the decision categories.

4. Applying risk & decision colors on pkg upload tab
> It might be a pain for the score, but will be super easy for the decision.

Issue #477, opened 2023-03-28
Update branch for chromote


AARON-CLARK:
According to @Jeff-Thompson12, the `chromote` fix has been moved to HEAD. We should probably update `renv.lock` so it's not pointing at some obscure branch that may get trimmed.

![image](https://user-images.githubusercontent.com/17878953/228305682-5cb6ca56-8b92-4051-89f4-8dc07e69565a.png)

Jeff-Thompson12:
I looked into this. The updated version has not been pushed the CRAN, but is on the main branch for `chromote`. Do we want to wait for the CRAN version?

PR #511, opened 2023-04-28
Update Chromote

Jeff-Thompson12:
Addresses #477 

Use the CRAN version for `{chromote}`


Robert-Krajcik (approved):
Looks OK to me.
Time to move off the `origin-testfix` version of `{chromote}`.

PR #513, opened 2023-05-04
Use Local Library

Jeff-Thompson12:
This PR sets to resolve some issues we have been encountering (like where's the documentation) and setup the application for new features coming up.


Jeff-Thompson12:
I have made a change that was probably needed a while ago in this PR, and that is to use `glue::glue_sql()` within the function `dbSelect()` and `dbUpdate()`. Half of the time we were passing a glued string anyways. And there were instances of it failing (e.g. try uploading `cli`).

Jeff-Thompson12:
Items still to do:

- [ ] Initializing application with empty library
- [ ] Checks on verifying user library
- [ ] Handling reweighting. What if package is not in library? What if the version does not match?
- [ ] Do we want more user control over the library in the application itself?
- [ ] Tests! Our current tests completely bypass this interactivity. How do we make sure everything is working correctly?

AARON-CLARK:
Hi @Jeff-Thompson12, just a quick note, given some [pharmaR/pkg_explorer](https://github.com/pharmaR/pkg_explorer/tree/main) code has been shared recently. It looks like that app won't see any of the package root dir files unless they are installed from source with some special options according to this SO answer. However, I haven't been able to reproduce this successfully. Just thought I'd share my progress (or lack of progress) so far.

```r
install.packages("Tplyr", type = "source", INSTALL_opts = "--with-keep.source")
```

https://stackoverflow.com/questions/48934886/how-do-i-keep-source-files-when-using-rs-devtools-library-function-install

Jeff-Thompson12:
@AARON-CLARK After several headaches I think this PR/issue needs to pivot. I think that instead of having a library being stored with the application, we should be storing `tar.gz` files. Here are the reasons why:

1. Different OS handle the R package installation process differently. It will be hard to standardize the process within the application and tell users how to manager their own libraries. By using `tar.gz`, we are using a standard R package distributed file that any savvy R user should know how to create/get.
2. We really want to use the package in two different ways. `riskmetric` seems to work best with installed packages. The package source does not necessarily have documentation because this can be compiled at installation. We really want to look at the installed package. We can easily install to a temporary library whenever we are creating the metrics if we have the `tar.gz` files.
3. We also want the source code. We want users to be able to browse the file directory. We can easily `untar` the `tar.gz` file to a temporary directory for this purpose.
4. The `tar.gz` files have the version number with them for a package. This allows us to also reach the goal of differentiating between versions.

Jeff-Thompson12:
Main issue I'm encountering with the `tar.gz` file approach is that you might need dependencies installed as well. When running `install.packages()` from CRAN, the dependencies will automatically be installed as well.

PR #514, opened 2023-05-05
Fix button alignment issues

Jeff-Thompson12:
Account for more scenarios that can cause the upload or remove package buttons to become mis-aligned with the selectize inputs.


AARON-CLARK (approved):
This looks good to me! I just merged with `dev` and updated the dev version & news. We'll wait for checks to pass, then merge.

Issue #520, opened 2023-05-08
Call `glue::glue_sql()` inside of `dbSelect()` and `dbUpdate()`


Jeff-Thompson12:
Most of the time we are passing arguments to these function with `glue::glue()` anyways. By passing `'` for the character values we are running into issues (e.g. try uploading package `cli`).

AARON-CLARK:
closed with #521

PR #521, opened 2023-05-08
Call `glue::glue_sql()` in `dbSelect()` and `dbUpdate()`

Jeff-Thompson12:
Addresses #520


AARON-CLARK (commented):
All db activity appears to work for me in my testing. Would appreciate another pair of eyes taking a look though. Maybe @Robert-Krajcik or @imran3004? Thanks for the PR!

Jeff-Thompson12:
Small edit to implement this new logic on the summary comments added with #512.

Also updated NEWS and version number.

Issue #523, opened 2023-05-17
Apostrophes in decision categories break things


Jeff-Thompson12:
I noticed if there is an apostrophe in a decision category name, everything falls apart, which is surprising because I thought we switched over to `glue::glue_sql()`:

```yml
decisions:
    categories:
      - Super GxP Compliant
      - It's okay
      - Pending Review
      - Not Promising
      - Terrible pkg
```

Console:

![image](https://github.com/pharmaR/riskassessment/assets/17878953/7ffe6e9e-7573-4562-a5f8-8524f444bb22)

App:

![image](https://github.com/pharmaR/riskassessment/assets/17878953/480c3877-a645-44fa-a430-50e8da8ea3d8)

AARON-CLARK:
Related to #520

AARON-CLARK:
closed with #524

PR #524, opened 2023-05-17
Fix apostrophe issue with decision categories

Jeff-Thompson12:
Addresses #523


Jeff-Thompson12:
The issue wasn't just inserting the category into the database table, but since id's were being dynamically created, bad values like apostrophes would break UI elements as well.

AARON-CLARK (approved):
Looks really good to me 👍🏼 

image

image

Issue #346, opened 2023-01-13
Expand User Roles


Robert-Krajcik:
We need user roles going beyond `admin` versus `nonadmin` 

Maybe ditch `nonadmin` in favor of other roles, like `reviewer`, `user`, `approver`, ???
What privileges would each role assume?

AARON-CLARK:
Thanks @Robert-Krajcik, I agree - it's time we consider creating some roles besides `admin` and `nonadmin`. I could see an organization wanting these roles, for example:


`admin` - all privileges
`lead` - all privileges of an admin, minus the ability to manage userids/passwords. So basically, they still have full access to the assessment re-weighting tab. I like the name `lead` over "approver" only because their job may include rejecting pkgs too. So in that way, they do more than approving.
`reviewer` - all privileges of a lead, except ...

* can't make any final package decisions
* can't revert package decisions
* can't delete packages from db (feature coming soon. See #340 )
* no access to anything from administrative tools tab

Also, I'm open to better naming options for these roles.

AARON-CLARK:
With the recent addition of #487, perhaps certain roles or users have access to review certain classes of packages. For example, perhaps I want to create a role that can be a `reviewer` on any package but has `lead` type privilege for "TLG" class packages!

AARON-CLARK:
closed via #527

PR #527, opened 2023-05-25
Expand User Roles

Jeff-Thompson12:
More fully addresses #346


Jeff-Thompson12:
Few notes to keep in mind during review:

1. May have forgotten some categories.
2. I implemented a tier system along with individual privileges
3. The default configuration has the tier system. The example configuration has individualized privileges. A configuration can really have both though.

AARON-CLARK:
> @AARON-CLARK How do you see the overall comment in the sidebar functioning differently than the package summary in the build report? If the overall comment is meant to be an explanation of the decision made, should we move the initial text input for the comment to the submit decision module?

The purpose of `Overall comment` is up for debate, but right now it's meant to serve as a short hand comment that can focus on anything the user wants to, although it is un-editable after the decision has been made and we also used use it to record important messages from decision automation, like so:

image

We did decide not to add risk re-calculation comments there - which only further emphasizes it's ties to being more of a "decision comment". The risk re-calc comments show up on the metric tabs as seen below:

image

Maybe it would be best to re-label `Overall comment` to `Decision Comment`? And if that were the case, I think it'd make more sense to live in the submit decision module.

What do you think?

Jeff-Thompson12:
@AARON-CLARK I think we need to transition it to be a decision comment and add a "general" comment that functions like the Maintenance or Community metric comments, where multiple comments can be submitted by multiple users.

We may want to allow comments to be edited or deleted as well. Although that can snow ball quickly.

AARON-CLARK:
> I think we need to transition it to be a decision comment and add a "general" comment that functions like the Maintenance or Community metric comments, where multiple comments can be submitted by multiple users.

Yeah, I think it makes even more sense to generalize the mm comments & cum comments into a general comment when we join those two tabs together in the near future. Maybe we create a subject line for the comment? So if users want to comment on "Testing" they could separate their comments into separate thoughts, as they have them or review that section of the app. Do we control the subjects or are they user-defined? Or maybe no subjects.

Also, does this work pertain to this PR? lol

Jeff-Thompson12:
Don't be hating. I was thinking about roles/privileges to make certain kinds of comments and it took off from there.

AARON-CLARK:
I gave this PR a quick initial review & it looks good so far, but I think that the idea of "tiers" is a little confusing because (1) they seem to be hardcoded with certain privledges & it's not obvious what those are (without a vignette) and (2) aren't the roles themselves supposed to represent groupings of different privledges? So I'm not sure what value the "tiers layer" is adding.

Let me explain: what if the categories under the `credentials` yaml were organized a bit differently? In the example below, I envision the `roles:` to stay the same but for `privledges:` to give those roles groups or single abilities, not the other way around (giving the abilities to one or several roles). I'm only suggesting this because it allows app-deployers to define their own custom tiers for each role, without the messiness of a `tier1` & `tier2` tiered system, which seems duplicative. Would something like this work?

```yml
credentials:
    roles:
      - admin
      - lead
      - reviewer
      - SME
    privileges:
      admin: admin
      lead: [weight_adjust, decision_adjust, final_decision, revert_decision, add_package, delete_package]
      reviewer: add_package
      SME: [add_package, decision_adjust, revert_decision]
```

In the app, if the `admin` was going to create a new user, they should see non-admin options for lead, reviewer, SME, and then all the individual privledges, namely: weight_adjust, decision_adjust, final_decision, revert_decision, add_package, delete_package.

image

When checking if a certain user has access to a feature in the app, you'd have to just look up the users role & that role's privledges.

Thoughts? Criticisms? Critiques?

Jeff-Thompson12:
> Thoughts? Criticisms? Critiques?

Okay, I like this framework a lot more than what I put together as far as defining the roles in the yaml. I will remove the tier system. I was trying to simplify some logic, but it didn't really work out as intended.

I did overlook displaying roles (and maybe the ability to edit them) within the application itself. Do you envision privileges being connected to individuals ultimately with the role pre-populating those values?

AARON-CLARK:
> Do you envision privileges being connected to individuals ultimately with the role pre-populating those values?

I envision privileges being connected to roles, which should pre-populate those privledge values for individuals, but that brings up a good point: what if an admin wants to create a custom role on the fly with 2 or more privileges without re-deploying the app? I don't think we want to make a module to "define roles" in the app, do we? I'm not sure how well such a module would integrate with `shinymanager`? But perhaps a compromise could be to make the role/privledge selector a multi-select? Do we have the ability to change that dropdown? Doing so **would** shift privileges to be connected to individuals.

Jeff-Thompson12:
> I envision privileges being connected to roles, which should pre-populate those privledge values for individuals, but that brings up a good point: what if an admin wants to create a custom role on the fly with 2 or more privileges without re-deploying the app? I don't think we want to make a module to "define roles" in the app, do we? I'm not sure how well such a module would integrate with `shinymanager`? But perhaps a compromise could be to make the role/privledge selector a multi-select? Do we have the ability to change that dropdown? Doing so **would** shift privileges to be connected to individuals.

Yikes, I don't think having a multiple input field would mesh well with `shinymanager`. What I do think we can do is have an informative tab within the administrative tools. Maybe we can decide later on if we want to allow admins to create custom roles or edit existing ones. I certainly don't want to implement that functionality in this PR though.

Jeff-Thompson12:
@AARON-CLARK we are ready for another look here.

AARON-CLARK (changes requested):
I love the "Roles & Privileges" tab. Looks really good! Couple of things I'm noticing right off the bat is:

## (1) The single privledge option is only allowed while the `admin` checkbox is marked

image

None of those options are available when unchecked. I assume we'd probably want the opposite, right? The only time we'd want single-privledge options available is when the user is NOT an `Admin`.

## (2) I'm testing things the old fashioned way (using `credentials.sqlite` instead of `pre_auth_user` & `credentials_dev.sqlite`) and there are some strange behaviors:

First, I logged in as an `ADMIN` initially using the default `QWERTY1` password and it looks like I can't add packages, which is weird because I'm an admin. The widgets below have been disabled:

image

I'm thinking the reason is because we force users to keep "admin" all caps with initial log in, and there privileges are tied to case sensitive usernames? Although that doesn't makes sense because I'm clearly listed as an "admin" user below. It should be tied to my role
Next, I created the following users and then logged out to test them all:

image

When I signed is as "leadonly", things seemed to work as expected. I could add a pkg, make a decision, revert the decision, re-weight metrics, delete package. Without looking through the code, what does "decision adjust" do?

Also, I tried adding a user with no privileges which is a solid use case. I think they'd just be able to look at stuff, comment, and download reports. That seemed to work well.

Jeff-Thompson12:
@AARON-CLARK ready for another look

Jeff-Thompson12:
> I ran the app with default config, but wanted to check out the "lead" user with "lead" role. To do this, all I need to do is run `run_app(pre_auth_user = "lead")`, correct? Well, I did that and it looks like I had all the privileges outlined in the config, except for "final_decision" and "revert_decision". Could you check on that?

I think a bug got introduced her from another PR. Will implement the fix.

AARON-CLARK (approved):
Looks good, just needs a bullet point added to the news file & dev version incremented. Thanks for all your hard work @Jeff-Thompson12!

Issue #515, opened 2023-05-05
Confusing messaging on pkg upload progress bar


AARON-CLARK:
![image](https://user-images.githubusercontent.com/17878953/236542368-484b9bcd-71d5-446b-b485-7372ed446aa7.png)

The messaging above really only made sense when folks were uploading pkgs via a CSV file and the pkg version didn't align with the current version available on CRAN. Now that we can upload pkgs with a `selectizeInput()` or automatically via the new `Dependencies` tab, this lingo doesn't make sense here. Would be great if we only dropped the "...not '0.0.0'" phrase in these cases.

Jeff-Thompson12:
@AARON-CLARK This will be resolved when #513 is finished, but I can understand doing a small patch PR.

imran3004:
Closing the issue as the PR is merged to `dev` branch.

PR #528, opened 2023-05-29
Fix the confusing message on pkg upload progress bar (issue #515).

imran3004:
Added a check for packages being uploaded from the drop-down or read from CSV file.

The message about different versions is given only if the package is read from the CSV file and the version on CRAN is different from the version specified in the CSV file.


Xyarz (approved):
looks good to me!

Issue #482, opened 2023-04-10
Initialize metric weights from config file


AARON-CLARK:
Right now, metric weights are initialized from a sql table, but it would make more sense for the sql table to be initialized with metric weights stored in the config file. If metric name not included in the config, it should received default weight of 1. To remove a metric, it should receive weight of 0.

Jeff-Thompson12:
@AARON-CLARK Currently we are forcing the weight for `covr_coverage` to be 0 in the app. Do we want to change this? Or should this PR force that value to be 0 as well?

AARON-CLARK:
closed with #529

PR #529, opened 2023-06-01
Metric Weight Configuration

Jeff-Thompson12:
Addresses #482

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -17,6 +17,7 @@
 * Enhanced the "Database View" tab to include the decision category table
 * Allowed the user to set a custom color for decision categories (#465)
 * Add a check to give message about different versions only when package name is given from CSV file (#515)
+* Updated the version of `{riskmetric}` used to V0.2.2
 * Allow metric weights to be initialized in the configuration file (#482)
 
 # riskassessment 0.1.1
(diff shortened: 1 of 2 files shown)


Jeff-Thompson12:
I'm not entirely sure why the tests were failing. It does not seem like it should have been related to this PR, but here we are.

AARON-CLARK (commented):
I can't find anything wrong with this PR, which means I must be slipping. I'll wait for @borgmaan & @Xyarz to give it a look.

I think once this PR is merged, we'll have to include a new card for the covr_coverage metric. We've kind of been ignoring it for a long time, but just adding a "Percentage card" similar to bug closure rate (seen below) with updated labels should do the trick. I'll create a new issue.

image

Xyarz (approved):
Looks also good to me. Can't find anything either. Tested and everything worked as intended,

AARON-CLARK (changes requested):
@Jeff-Thompson12, you composed a perfect PR. @Xyarz & I couldn't find anything wrong with it. Congratulations!

Could you bump the version number & updates the `NEWS` and we'll get this one merged!

Jeff-Thompson12:
I'm just sad that neither one of you mentioned the inclusion of the new reactive `trigger_events` being passed between modules now that will enable communication between different parts of the application without making modules directly connected to each other.

Jeff-Thompson12:
Nevermind, that was one of my imperfect PRs. They blend together for me.

Issue #525, opened 2023-05-18
Metric Re-weighting bugs


AARON-CLARK:
First, if there are no packages in the db, you shouldn't be able to re-weight. If you click the button, the app will crash.

Second, if you just uploaded pkgs and then re-weight, we should clear out the uploaded pkgs summary section, as those risk scores will conflict with the newly re-weighted risk score.

![image](https://github.com/pharmaR/riskassessment/assets/17878953/4c10e87c-0378-4082-ad03-8688d81b7424)

AARON-CLARK:
closed with #530

PR #530, opened 2023-06-01
Re-weighting Bugs

Jeff-Thompson12:
Addresses #525

### DIFF
--- a/R/mod_reweightView.R
+++ b/R/mod_reweightView.R
@@ -15,6 +15,7 @@ reweightViewUI <- function(id) {
 #' @param id the module id
 #' @param user the user name
 #' @param decision_list the list containing the decision automation criteria
+#' @param trigger_events a reactive values object to trigger actions here or elsewhere
 #' 
 #' 
 #' @import dplyr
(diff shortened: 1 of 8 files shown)


R/mod_reweightView.R
```diff
@@ -273,8 +276,17 @@ reweightViewServer <- function(id, user, decision_list) {
...
+      curr_new_wts(
+        get_metric_weights() %>%
+          dplyr::mutate(new_weight = weight) %>%
+          dplyr::mutate(weight = ifelse(name == "covr_coverage", 0, weight)))
+      
+      user$metrics_reweighted <- user$metrics_reweighted + 1
+      
       # update for each package
       all_pkgs <- dbSelect("SELECT DISTINCT name AS pkg_name FROM package")
+      req(nrow(all_pkgs) > 0)
```
borgmaan, line 288: I would at least add a comment here saying why this is important and that deleting it will re-introduce a bug.

A unit test would make sense as well, but I can't recall where the dev team landed on the testing strategy (e.g. did we adopt the mantra that we add tests for 1) new functionality and 2) any bugs that are discovered and fixed to prevent regression?).

borgmaan (changes requested):
One small change. Otherwise looking good.

R/mod_reweightView.R
Jeff-Thompson12: Both good points. I will add both.

borgmaan (approved):
👍

R/mod_reweightView.R
AARON-CLARK: > did we adopt the mantra that we add tests for 1) new functionality and 2) any bugs that are discovered and fixed to prevent regression?).

I think we landed on #1 for sure. But I like #2 also. Let's do both. I'll add it to our PR checklist wiki.

PR #531, opened 2023-06-02
Update `{riskmetric}`

Jeff-Thompson12:
Update the version of `{riskmetric}` being used in the app

### DIFF
--- a/renv.lock
+++ b/renv.lock
@@ -1795,15 +1795,15 @@
     },
     "riskmetric": {
       "Package": "riskmetric",
-      "Version": "0.2.0",
+      "Version": "0.2.2",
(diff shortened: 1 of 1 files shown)


Jeff-Thompson12:
@AARON-CLARK I also realized that we were suing an older version of `{riskmetric}`, so here is a PR to update it.

AARON-CLARK (approved):
NOICE

Jeff-Thompson12:
@AARON-CLARK These test failures kind of started popping up out of no where. I have a fix on #529, but this PR isn't the reason for the failures.

AARON-CLARK (approved):
👍🏼

Issue #361, opened 2023-01-20
Audit Table(s) in Database


borgmaan:
It looks like there has been discussion and development around creating an audit trail for user activity in #134.

Has the addition of adding audit logging capabilities to the database been discussed?

This can be achieved pretty quickly by adding some triggers to the existing tables that log info to an audit table about what changes were made during a given transaction (e.g. what table what changed, new & old values, time of change, etc.).  

 In addition to #244, I think this is something that organizations will look for when assessing the stability of the product.

Jeff-Thompson12:
I'm starting to look into this idea (which I really like). Here are the challenges I'm facing:

- [ ] We have processes that are delete/insert instead of update (see reweighting). This poses a problem with tracking.
- [ ] SQLite does not use variables, so we will have to have some way to track who is making the changes. This would be straightforward except that some changes are done by the system (such as auto decisions).
- [ ] It is unclear if any database re-structuring needs to take place (e.g. are our primary keys properly defined? Does it even matter for this process?)
- [ ] What are the implications with creating functions to update databases once this is implemented?
- [ ] How are we going to allow the audit tables to be viewed by administrators? Does this affect how we can store the logged information?
- [ ] Does this phase out using the logging file? If not, how do we see these serving different purposes?

PR #535, opened 2023-06-07
Audit DB Tables

Jeff-Thompson12:
Addresses #361

### DIFF
--- a/R/utils_startup.R
+++ b/R/utils_startup.R
@@ -25,7 +25,6 @@ create_db <- function(db_name){
     "create_decision_table.sql",
     "create_package_table.sql",
     "create_metric_table.sql",
-    "initialize_metric_table.sql",
     "create_package_metrics_table.sql",
     "create_community_usage_metrics_table.sql",
     "create_comments_table.sql"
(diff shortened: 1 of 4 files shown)


Jeff-Thompson12:
@pharmaR/riskassessment I would appreciate any insight on what is going on with the R CMD check for this branch. The tests pass locally running `devtools::test()` but fail when running `devtools::check()`. It always fails at the same spot ([after submitting the decision rules](https://github.com/pharmaR/riskassessment/blob/86ee2ebff32fa9cbf23811db78a16acdab384708/tests/testthat/test-decision_automation.R#L63)). The [changed lines](https://github.com/pharmaR/riskassessment/blob/86ee2ebff32fa9cbf23811db78a16acdab384708/R/mod_decision_automation.R#L545-L553) corresponding to the failure are also not very informative. They appear to only affect the timing of the process, but increasing the timeout does not seem to work. Plus the R CMD check seems off for it just being a waiting issue (i.e. "callr subprocess failed: could not start R, exited with non-zero status, has crashed or was killed").

Note the following that I have tried:
- Rewriting the `dbUpdate()` into a `for` loop
- Using a version of the test database that does not include the audit log triggers
- Adjusting the `app$wait_for_idle()` to `app$wait_for_value()` with extended timeout
- Also note the checks pass with using a database including the audit log triggers and reverting the [changed lines](https://github.com/pharmaR/riskassessment/blob/86ee2ebff32fa9cbf23811db78a16acdab384708/R/mod_decision_automation.R#L545-L553) back to what is currently on the `dev` branch.

Issue #478, opened 2023-03-28
Update `Deployments.Rmd` to reflect new config file usage


AARON-CLARK:
Following merge of #459, the deployments vignette will need a somewhat large overhaul, describing how to use the config file. This certainly needs to be done before our next release. The vignette should cover these config toggles:

- [x] assessment_db
- [x] credential_db
- [x] loggit_json
- [x] credentials
    - [x] roles
    - [x] privileges
- [x] decisions
    - [x] categories
    - [x] rules
    - [x] colors
- [x] metric_weights

AARON-CLARK:
closed with #555 & #542

PR #542, opened 2023-06-13
Update User Guides

Jeff-Thompson12:
Addresses #478


AARON-CLARK (approved):
You took the words right out of my mouth

Issue #549, opened 2023-06-15
Checks failing on macOS


Jeff-Thompson12:
Has anyone else noticed the errors for macOS lately?

<img width="100%" alt="Snag_af88293" src="https://github.com/pharmaR/riskassessment/assets/85879620/a9bc9058-a4e7-45d6-8680-7b8bdd33821c">

imran3004:
@Jeff-Thompson12 , what are the steps to reproduce this please?

Jeff-Thompson12:
@imran3004 It is on the [Github Actions for the Next Release PR](https://github.com/pharmaR/riskassessment/actions/runs/5247682369/jobs/9478159444?pr=472).

Jeff-Thompson12:
I think I am slowing figuring out what is occurring here. It appears that providing the repository in `renv.lock` does not dictate how the packages are being installed in `renv::restore()`. When doing the restore, `{renv}` utilizes the HASH.

You can further notice that we are using versions of packages that are older than the RSPM snapshot date. It probably makes sense to reset the project based on that date.

Jeff-Thompson12:
Some of the HASHs in the `renv.lock` file are pointing to the microsoft.mran portal which is scheduled to close soon. I believe this to mean that users will not be able to initialize their project with our lock file unless we update the packages to point to the RSPM url.

Jeff-Thompson12:
Note the retirement of MRAN [here](https://techcommunity.microsoft.com/t5/azure-sql-blog/microsoft-r-application-network-retirement/ba-p/3707161)

Jeff-Thompson12:
The issue boils down to this synopsis: 

>The Microsoft R Application Network (MRAN) is being retired on July 1, 2023. When running `renv::restore()`, the package and version from `renv.lock` is searched for on MRAN. This is done because MRAN was storing the binaries for different operating systems. By disabling MRAN, `renv::restore()` will use other means (like the repo specified in the lock file), but will try to install these packages from source. This is causing the GHAs to fail during the setup process. The only way to fix this is to figure out how to configure the virtual machine so that the packages can be installed from source. This only gives us a benefit if we anticipate both that these operating systems will be used for deployment of the application and that there is some functional difference between OS that require us to account for that different behavior. The second is unlikely because we are not compiling any code for use in the application.

It is probably beneficial for us to disable using MRAN, since an older version of `{renv}` is being utilized in the project. But we should probably consider removing the R-CMD checks for Windows and Mac.

[...]

PR #556, opened 2023-06-22
Fix GHA macOS

Jeff-Thompson12:
Addresses #549


Jeff-Thompson12:
Deemed waste of time. Connect with @borgmaan with inquiries.

Issue #564, opened 2023-06-27
"Viewer" role & "comment" privilege


AARON-CLARK:
I'm thinking about the prospect of having a user role where they can't do anything. Literally, all they can do is view/ observe the deployed app, navigating from tab to tab, but perhaps we don't want them to even comment. If that's the case, we should add "comment" as a new privilege where we could give users access to the comment modules located around the application.

I tested the following, and it looks like leaving a role's privilege bucket empty in the config file works well!
```yml
default:
  assessment_db: database.sqlite
  loggit_json: loggit.json
  credential_db: credentials.sqlite
  credentials:
    roles:
      - admin
      - lead
      - viewer
    privileges:
      admin: [admin, weight_adjust, auto_decision_adjust, final_decision, revert_decision, add_package, delete_package]
      lead: [weight_adjust, auto_decision_adjust, final_decision, revert_decision, add_package, delete_package]
      viewer: 
  decisions:
```

AARON-CLARK:
closed with #565

PR #565, opened 2023-06-28
Add commentator privileges

Jeff-Thompson12:
Addresses #564 

Also adds a warning if any privileges are not assigned to any roles in the application.


AARON-CLARK (changes requested):
@Jeff-Thompson12, nice work on this PR. I haven't run any code yet, but just from looking at the diff, I had a few suggestions below.

However, first I have a comment/question: I feel like the `overall_comment` privilege (access to overall comment and package summary UIs) should be added to the default privileges for the `reviewer` role. That way, a reviewer could be seen as the person doing most of the "grunt work". They are the person tasked with uploading a package, writing the report, and creating a short synopsis regarding a final decision. Then the "lead" can evaluate the review and make a final decision. What do you think?

Requesting changes for the reasons below:

(1) We should probably move away from using images to display the contents of the config file, since it is harder to update. Instead we should just use the triple back tick and `yml` keyword to generate these code blurbs.

image

(2) This vignette is weak sauce compared to the other privileges! 😆 

image

Jeff-Thompson12:
> (2) This vignette is weak sauce compared to the other privileges! 😆

Is there a requested change here, or are you just mocking the low effort put in updating the vignette?

AARON-CLARK:
> > (2) This vignette is weak sauce compared to the other privileges! 😆
> 
> Is there a requested change here, or are you just mocking the low effort put in updating the vignette?

Both? Could you please add an image or two to show off the UI elements impacted by these privileges? Or something along those lines. Perhaps even elaborate on why the two comment privileges are separate?

vignettes/User_Roles_and_Privileges.Rmd
```diff
@@ -305,13 +305,21 @@ A user with the `delete_package` privilege can remove packages from consideratio
 
 ### Overall Commentator
 
-A user with the `overall_comment` privilege can add/edit the overall comment and summary of a package.
+A user with the `overall_comment` privilege can add/edit the overall comment and summary of a package. Unlike maintenance or community usage metrics, each package can only have one overall comment and summary. A user with the `overall_comment` privilege can provide the overall recommendation for the package decision and provide the summary of what the package offers.
+
```
AARON-CLARK, line 309: Add a <br> here to create a little spacing between the paragraph and the image. But if you do that, you may want to do it preceding each image on this page/ vignette.

vignettes/images/start_decision_automation.png
AARON-CLARK, line 1: You didn't like my arrow?

AARON-CLARK (changes requested):
Hi @Jeff-Thompson12, looks good. I have a few comments/ suggestions below.

vignettes/images/start_decision_automation.png
Jeff-Thompson12: Sorry. Was hoping to sneak that change in.

vignettes/User_Roles_and_Privileges.Rmd
Jeff-Thompson12: I added a <br> before and after images.

AARON-CLARK (approved):
👍🏼

PR #569, opened 2023-07-06
`covr-coverage` fixes

AARON-CLARK:


inst/sql_queries/initialize_metric_table.sql
```diff
@@ -11,5 +11,5 @@ VALUES
 ('export_help',         'Documentation',     '% of documented objects',          1, 0, 'maintenance', 1),
 ('bugs_status',         'Bugs Closure Rate', '% of the last 30 bugs closed',     1, 0, 'maintenance', 1),
 ('license',             'License',           "Package's license",                0, 0, 'maintenance', 1),
-('covr_coverage',       'Test Coverage',     '% of objects tested',     1, 1, 'maintenance', 0),
+('covr_coverage',       'Test Coverage',     '% of objects tested',     1, 1, 'maintenance', 1),
```
Jeff-Thompson12, line 14:
```suggestion
('covr_coverage',       'Test Coverage',     '% of objects tested',     1, 0, 'maintenance', 1),
```

Since we are already here, we should probably also mark that it is not a URL.

Jeff-Thompson12 (approved):

PR #570, opened 2023-07-12
Upgrade `renv`

Jeff-Thompson12:
Opening a draft PR to upgrade `renv`.

### DIFF
--- a/renv.lock
+++ b/renv.lock
@@ -8,37 +8,43 @@
       }
     ]
   },
+  "Bioconductor": {
+    "Version": "3.16"
+  },
   "Packages": {
     "AsioHeaders": {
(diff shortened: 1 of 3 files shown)


Jeff-Thompson12:
When upgrading `renv`, it updated the `renv.lock` file based on my library. I believe the only package (besides `renv` itself) that has gotten changed is `chromote`. I have been using the developmental version from Github because of the issues with being able to view the app during debugging.

Jeff-Thompson12:
I have added the "Urgent" label to this PR since the GHAs are currently failing with using version 0.16.0.

AARON-CLARK (approved):
LGTM

Issue #438, opened 2023-02-23
New Comm Card Proposal: Download Trend


AARON-CLARK:
I want to propose a new card that fits a linear model to the monthly data produced by `generate_comm_data()` with the goal of reporting on the slope coefficient.  For example, below you can see that the `stringr` package has been increasing in downloads over the last two years. Let's say we generate a linear model on this data and the slope is 27k. In this context, we could highlight two things:

* First, the the trend is positive/ upwards!
* Second, that the average increase in pkg downloads increases by 27k per month.

I think those are really handy statistics to know! Maybe we add a button to overlay the linear model on the plot? We probably want to limit the model fit to the last two years to keep the stats recent... but I guess we could always empower the user to choose the timeframe. On the lower bound, we shouldn't fit a model unless a minimum of 12 months exists.

Perhaps if we implement and test some scenarios, this could be an assessment we propose to `riskmetric`?

Thoughts?

This current Community Usage Tab:

![image](https://user-images.githubusercontent.com/17878953/220915732-10577a8b-65c9-4ad0-8acc-739d6273722b.png)

Eduardodudu:
Hi!

I have added a first approach for the card, for now, the timeframe is strict for 12 months and the card is not shown when it does not fit the minimum of 12 months.


![Image](https://github.com/pharmaR/riskassessment/assets/8216229/7d0a17c2-a4da-46eb-a078-996afab2106e)

![Image](https://github.com/pharmaR/riskassessment/assets/8216229/ce265071-cf5c-4354-9b78-de3f122a22ec)


For the plot, I added the linear trend of 2 years, since plotly was being used straight without ggplot, I couldn't replicate geom_smooth and had to create the lm + predict approach:



![Image](https://github.com/pharmaR/riskassessment/assets/8216229/6a591da9-1521-4031-9d18-e372b4d83e00)


I also added legends at the bottom

Let me know what you think about it, there is a draft pull request of this task.

AARON-CLARK:
closed with #572

PR #572, opened 2023-07-13
New Comm Card Proposal: Download Trend #438

Eduardodudu:
## TASK

Feature task on the linear trend card and plot information #438 

## This PR adds:

1. A first approach for the card, for now, the timeframe is strict for 12 months, and the card is not shown when it does not fit the minimum of 12 months.

Image

Image

2. For the plot: A linear trend of 2 years

Image

Also added legends at the bottom


R/utils.R
```diff
@@ -365,6 +365,39 @@ build_comm_cards <- function(data){
...
+      TRUE ~ "bars"
+      )
+    ) %>%
+    dplyr::select(estimate, succ_icon)
+  
+  cards <- cards %>%
+    dplyr::add_row(
+      name = 'downloads_trend',
+      title = 'Downloads trend',
+      desc = 'Trend of downloads in last 12 months',
```
AARON-CLARK, line 393: Technically, the trend data could be fit on anywhere from 12 to 24 months worth of data, correct? I think it'd be easier to just leave the time period out of the description.  If you want to take the easy route, I think this description more accurately conveys the analysis, no?

```suggestion
      desc = 'Trend in average downloads per month',
```
Or "Monthly trend in downloads, on average" would work too. 

Or you could insert a dynamic time period- one that changes from package to package. For example a newer package may be fit to 14 months of data if the first release was 14 months ago, but `stringr` would always be fit to 24 months of data since it's been around a long time.

R/utils.R
```diff
@@ -487,22 +520,42 @@ build_comm_plotly <- function(data = NULL, pkg_name = NULL) {
...
-           margin = list(t = 100),
-           showlegend = FALSE,
-           yaxis = list(title = "Downloads"),
-           xaxis = list(title = "", type = 'date', tickformat = "%b %Y",
-                        range = dates_range)) %>% 
+
+  #Create linear model
+  data_horizon <- downloads_data %>% 
+    dplyr::arrange(day_month_year) %>%
+    filter(day_month_year >= max(day_month_year) - lubridate::years(2)) %>%
```
AARON-CLARK, line 527: add `dplyr::` prefix

R/utils.R
```diff
@@ -365,6 +365,39 @@ build_comm_cards <- function(data){
...
+    return(cards)
+  }
+  
+  trend_downloads <- dplyr::as_tibble(data) %>% 
+    dplyr::arrange(year, month) %>%
+    dplyr::filter(row_number() >= (n() - 11)) %>% 
+    dplyr::mutate(row_n = row_number())
+  
+  model_result <- lm(downloads ~ row_n, data = trend_downloads) %>% 
+    broom::tidy() %>%
```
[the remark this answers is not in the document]
Eduardodudu: Sure! Removed `broom::tidy` approach

R/utils.R
Eduardodudu: I really liked the idea of showing the number of months being considered for the analysis, it's ranging from 12-24 months and I explicitly state it in the card now

R/utils.R
Eduardodudu: Thanks!

Eduardodudu:
> Hi @Eduardodudu, great Draft PR! Thanks again for taking on this issue. I think it's a fun one. I do have a few notes to share.
> 
> First, I performed a test on a package called `tidyCDISC` below, which only has only existed on CRAN for 10 months, but it's still displaying a card & and fit line in the `plotly` object. Didn't you say that if there was < 12 months of data, the card or the fit line wouldn't show? I think that is a good route to take.
> 
> image
> 
> I also checked out `rhino` as you did, and there is actually 16 months worth of data, so it should display a card with the trend, correct?
> 
> image
> 
> I've also jotted down a few additional comments below. Let me know if I misinterpreted something. Great start! 💪🏼

You're correct! It was causing an issue with the maximum range of 12 months, fixed and did a small refactor to not calculate objects unnecessarily when they're below 12 months and not being displayed. 

Let me know what you think!

R/utils.R
```diff
@@ -390,7 +397,7 @@ build_comm_cards <- function(data){
     dplyr::add_row(
       name = 'downloads_trend',
       title = 'Downloads trend',
-      desc = 'Trend of downloads in last 12 months',
+      desc = glue::glue("Trend of downloads in last {amount_months} months"),
```
AARON-CLARK, line 400: I still think we need to be clear that this is a "monthly" trend, so users know that this value reflects an average increase or decrease in downloads per month. Consider adding the word "Monthly"? Or some other phrase that fits better. I know real estate is tight.

AARON-CLARK (changes requested):
I think this is working as expected now @Eduardodudu! Thanks for again for the PR. I just have one additional comment.

@Jeff-Thompson12, do you think you could put a second pair of eyes on it to catch anything I missed? Thank you.

AARON-CLARK:
Hi @Eduardodudu, per our discussion today, we talked about making the downward trend icon and the model coefficient value red in the card when < 0.

In addition, we talked about changing the color of the fit line to be something that pops a little more - perhaps a dark orange, or some other color options you could dream up.

Thanks again!

R/utils.R
Eduardodudu: Thanks! Updated the title to 'Monthly downloads trend'

image

Eduardodudu:
> Hi @Eduardodudu, per our discussion today, we talked about making the downward trend icon and the model coefficient value red in the card when < 0.
> 
> In addition, we talked about changing the color of the fit line to be something that pops a little more - perhaps a dark orange, or some other color options you could dream up.
> 
> Thanks again!

Hi @AARON-CLARK!

Updated the behavior for the cards having red icon + value and the plot to dark orange

Let me know what you think about it

image

AARON-CLARK (changes requested):
@Eduardodudu, looks good to me. Thanks for taking the extra steps at the end to style the output. I think it looks a lot better.

Referencing the [PR checklist](https://github.com/pharmaR/riskassessment/wiki/Developer-Pull-Request-workflows#pr-checklist), I think the only thing you have left to do is increment the dev version using `usethis::use_version("dev")`.

Thanks!

Eduardodudu:
Thanks, @AARON-CLARK, updated!

AARON-CLARK (approved):
👍🏼

Issue #498, opened 2023-04-13
Add a "Package Explorer"


AARON-CLARK:
For context, GSK shared their pkg validation case study with the R Validation Hub in 2022, which featured a "Package Explorer" application that investigates:

1. Code for all exported functions (related to #242)
2. Exported Documentation (related to #259)
3. Tests (related to #260)

It's important to note that these features are important when a deeper dive (beyond the `riskmetric` assessments) are needed. Its especially useful for folks to browse the robustness of tests written for each function. Here is a link to their presentation, fast-forwarded to the good part:

  https://youtu.be/P2m6RmVViu0?t=2792

![image](https://user-images.githubusercontent.com/17878953/231810656-17d5a108-d2d9-46c1-bdf3-479b1c0fd381.png)

AARON-CLARK:
closed with #637

PR #574, opened 2023-07-14
Add Package Explorer

Jeff-Thompson12:
First step in addressing #498 by incorporating the package explorer code shared with team.


AARON-CLARK (changes requested):
Hi @Jeff-Thompson12,

Honestly, I think this looks really great already! I love the toggle-able Maint and Comm Usage metrics tab. For the Source Explorer, I pulled in `tidyCDISC` in since I'm the most familiar with that pkg. I didn't realize how much stuff is added to the .Rbuildignore file. One thing I wish was visible was the `.github` folder so users could check out the GHAs. But alas.

image

(1) On thing I'm wondering is if we should add a pkg name somewhere? Like perhaps below the "File Browser" header? Even though it's visible in the Package Control Panel. IDK, maybe it's not a good idea. We don't do it for Maint Metrics... Your thoughts?

(2) Also, I'm wondering if we could limit the length of the file tree so that it get's a scroll bar. That way, we don't have to scroll the entire page up and down just to change files.

(3) This is more of a big picture question, but is there any merit to making the package control panel collapsible to the left? Perhaps it would really alleviate some real-estate issues we have on these tabs. If it is collapsible, I'd want it to be really huge on the left hand side, making a "mega button" like below so that it's ultra-visible:

image

(4) A minor comment: do you think you could add a \<br\> or two at the bottom of these UIs so that there's a little margin between them and the "Checkout the App's code" banner?
image

(5) Also, this is extra credit (definitely not necessary), but what do you think about rendering html or markdown files (if they exist)? Once GSK releases their next batch of code to us, it will render all the `man` Rd files, but those only exist for exported functions. It would be kind of cool to render the README or NEWS. Anyway, just a thought. Not a requirement by any means.

(6) Last thought (sorry I have so many) is: should we make the whole process that creates the source explorer optional? For example, if people don't want to use a source explorer at all, we could give them that option to turn it off / on in the config file. If they don't want to use it, then we wouldn't have to download tarballs which will save time and computing power in the app, no? What are your thoughts? If you're for it, it could be tackled in a different PR.

Jeff-Thompson12:
> (1) On thing I'm wondering is if we should add a pkg name somewhere? Like perhaps below the "File Browser" header? Even though it's visible in the Package Control Panel. IDK, maybe it's not a good idea. We don't do it for Maint Metrics... Your thoughts?

Since we do not do this on any other tabs, I'm leaning towards no.
 
> (2) Also, I'm wondering if we could limit the length of the file tree so that it get's a scroll bar. That way, we don't have to scroll the entire page up and down just to change files.

I have made some edits to accommodate this.

> (3) This is more of a big picture question, but is there any merit to making the package control panel collapsible to the left? Perhaps it would really alleviate some real-estate issues we have on these tabs. If it is collapsible, I'd want it to be really huge on the left hand side, making a "mega button" like below so that it's ultra-visible:

We would need to look into this more. The big issue I see is our metric grids. All of the other tabs would need to look good as well if we changed their window size. Definitely outside the scope of this PR.

> (4) A minor comment: do you think you could add a <br> or two at the bottom of these UIs so that there's a little margin between them and the "Checkout the App's code" banner?

I have adjusted the padding to create more space between the main window and the footer.

> (5) Also, this is extra credit (definitely not necessary), but what do you think about rendering html or markdown files (if they exist)? Once GSK releases their next batch of code to us, it will render all the `man` Rd files, but those only exist for exported functions. It would be kind of cool to render the README or NEWS. Anyway, just a thought. Not a requirement by any means.

I would like to wait on this until we get the rest of their code. If they have already done some work with rendering, I would rather utilize/expand what they have than create something new for the time being.

> (6) Last thought (sorry I have so many) is: should we make the whole process that creates the source explorer optional? For example, if people don't want to use a source explorer at all, we could give them that option to turn it off / on in the config file. If they don't want to use it, then we wouldn't have to download tarballs which will save time and computing power in the app, no? What are your thoughts? If you're for it, it could be tackled in a different PR.

I have been leaning towards not storing tarballs, but utilizing the CRAN archive instead. I think this will resolve the issue of data storage. I would like to leave that to a separate PR though.

Issue #541, opened 2023-06-12
Add module to add / edit roles


AARON-CLARK:
Following merge of #527, this feature would allow `admin` users to edit or add roles in the app.
Would go nicely here:

![image](https://github.com/pharmaR/riskassessment/assets/17878953/f8e5f116-7686-444b-942a-b40deb08fbac)

PawanRamaMali:
@AARON-CLARK - Do you suggest the ability for `admin` users to change the role of other users belonging to any role type including `admin` to any other role types but not vice-versa? Also how do you visualize this change?

Jeff-Thompson12:
An `admin` user can already change the roles of other users. What is visualized here is managing the privileges associated with the defined roles (e.g. add the `delete_package` privilege to the `reviewer` role above) and being able to add or delete roles (e.g. add a new role `viewer` with no privileges). A few things (off the top of my head) would need to happen in order to implement this feature:

- [ ] Currently the roles and privileges are being populated by the configuration file for each session. We would need to change the behavior to be similar to other features where the configuration file initializes a database table(s) containing the role and privilege information. Much in the same way the decision category database table is created.
- [ ] The configuration checks for roles and privileges would need to be applied to the changes.
- [ ] Using a similar structure to other editing features would be desired (e.g. like the modal box that is used with the decision categories). We could have a gear icon button in the tab above that opens a dialog box for editing purposes.
- [ ] I also imagine that we might want some additional warnings in place. For example, a special warning could come up when adding or deleting the `admin` privilege. To me this is the hardest part because `{shinymanager}` has a special field for ADMIN. Any changes to the `admin` privilege would necessitate updating the credentials database itself to reflect those changes (e.g. if the `lead` role had the `admin` privilege and that privilege was subsequently removed, we would need to update the credentials database to "uncheck" ADMIN from all users who have the role of `lead`).

Jeff-Thompson12:
@PawanRamaMali Were you going to take a stab at this issue? I was thinking about working on it if not.

PawanRamaMali:
@Jeff-Thompson12 - If you're willing to work on it, I kindly request you to take the lead.

PR #576, opened 2023-07-18
Edit Roles & Privileges

Jeff-Thompson12:
Addresses #541


AARON-CLARK (changes requested):
Nice PR! The modal seems really intuitive. Couple of quick comments:

Could you re-label this to "Edit Role Name"? It seems that's the only action it's performing, no?

image

When deleting roles, shouldn't the lead, reviewer, and viewer roles be in this list? Or will we always keep the "core roles" that exist in the config file?

image

Things did start to get trippy when I started messing with the `admin` role. How about this scenario:

I logged on as an `admin` user and changed my role name from "admin" to something else. Then in the app, I no longer had permissions to submit decisions or overall comments! Even though my role isn't called `admin`, I still have the 'admin' privilege so shouldn't that cover me? If the role must be called "admin", then we shouldn't allow users to change the name of the admin role.

image

I did still have access to add packages, general comments, and access the re-weighting module. But I could NOT adjust auto decisions, final decisions, revert decisions, and overall comment. What's strange is even though I had the overall_comment disabled, I could still see and submit a pkg summary on the build report tab. Similarly, I could see the delete package `selectInput()` but I could not actually remove pkgs from the database.

FYI on the reweighting tab, I noticed the rows weren't highlighting anymore when changed to a new weight... not sure if this bug was introduced on this branch or not.

image

Jeff-Thompson12:
> Things did start to get trippy when I started messing with the `admin` role. How about this scenario:
> 
> I logged on as an `admin` user and changed my role name from "admin" to something else. Then in the app, I no longer had permissions to submit decisions or overall comments! Even though my role isn't called `admin`, I still have the 'admin' privilege so shouldn't that cover me? If the role must be called "admin", then we shouldn't allow users to change the name of the admin role.
> 
> image

The `admin` role is just privileges to manage credentials and is not a mega role that contains the subset of all other roles. If those options are unselected, you will no longer have access to them (even though you have full rights to re-implement them). I am unable to replicate any weird behavior with changing the name. Are you able to reproduce this behavior?

Jeff-Thompson12:
> I did still have access to add packages, general comments, and access the re-weighting module. But I could NOT adjust auto decisions, final decisions, revert decisions, and overall comment. What's strange is even though I had the overall_comment disabled, I could still see and submit a pkg summary on the build report tab. Similarly, I could see the delete package `selectInput()` but I could not actually remove pkgs from the database.

This is a good catch. The reactivity can be tricky here. I will have to check those rendering events and make sure they update with changes in the credentials.

Jeff-Thompson12:
> FYI on the reweighting tab, I noticed the rows weren't highlighting anymore when changed to a new weight... not sure if this bug was introduced on this branch or not.
> 
> image

Highlighting is working for me, but stops if I remove the reweighting privilege. This has to do with the issue you noted above where the UI of removed privileges is still visible (although not functional).

AARON-CLARK:
> I am unable to replicate any weird behavior with changing the name. Are you able to reproduce this behavior?

Yep. I tried it twice before submitting the review and a third time just now. Process: delete any credentials databases, loggin as pre-auth-user 'admin'. Change 'admin' role name to 'newName' and then head back over to `Risk Assessment` tab to find I can't submit or revert decisions... etc.

Jeff-Thompson12:
Did you also reinitialize `credentials_dev.sqlite`? What does the privileges table with the checkboxes look like?

Jeff-Thompson12:
Nevermind. I am able to replicate it now. Seems like the same issue with the reactivity. I will go through each privilege and make sure they are working properly.

Jeff-Thompson12:
@AARON-CLARK I'm pretty sure the reactivity is working in an expected way now. Before we only needed to concern ourselves with correctly displaying the app upon initialization. This PR added the possibility that an `admin` user might change their privileges mid session. I believe I have accounted for the various scenarios but let me know if you find anything.

AARON-CLARK (changes requested):
The reactivity works like a charm! Great job.

Only thing I can think to complain about would perhaps be with the "Delete Role" `selectInput()`. Specifically, I think it could get confusing if it has no values. Maybe consider changing the label to say "Delete unused roles" or insert some sort of tool tip or placeholder that communicates to the user that "No roles eligible for deletion. To delete a role, first ensure no users are assigned that role in the Credential Manager." for example.

image

Jeff-Thompson12:
Oops. Meant to add a blurb in about that. The select input is now disabled if empty and there is an information icon with a tooltip providing information about deleting roles.

image

Issue #579, opened 2023-07-18
File Browser crashes when switching pkgs and previewing certain files


AARON-CLARK:
The app crashes on `dev` with the following error when you try to open these files while browsing source data from the `{xts}` pkg:

```
Warning in file(con, "r") :
  cannot open file 'source/xts/R/ggplot2.zoo.R': No such file or directory
Warning: Error in file: cannot open the connection
```

![image](https://github.com/pharmaR/riskassessment/assets/17878953/9b18b839-5771-40a3-a6ef-aedd9830a35f)

AARON-CLARK:
Looks like it may be an issue with how the tree updates when switching from package-to-package. I just switched from `zoo` to `tidyCDISC` and some files update, like the `description` file seen here, but not the R/folder. Same is true for vignettes - they are all `zoo` vignettes. So when the app is told to go preview these files, it couldn't look them up because those files don't exist for `tidyCDISC`.

![image](https://github.com/pharmaR/riskassessment/assets/17878953/bc2a6280-68e6-4db3-a8d5-58abbbaf4324)

Jeff-Thompson12:
So I found this quite puzzling because I had checked this specific issue while integrating the code, but it appears that I introduced breaking changes in later commits. It appears that rendering the UI in the server is causing the tree to break.

PR #580, opened 2023-07-19
Fix rendering issue with jsTree

Jeff-Thompson12:
Addresses #579


Jeff-Thompson12:
@AARON-CLARK It appears that the rendering stopped working correctly when I started using `uiOutput()` for the tags. I have circumvented this issue by using a conditional panel. I think we should work towards using a different package to accomplish the file tree structure. The `{jsTreeR}` package does not behave in the expected way and may be hard to maintain the code surrounding it as package development proceeds. I will make an issue on their Github page though.

AARON-CLARK (approved):
looks good to me.

PR #591, opened 2023-07-20
Introduce DB stats via cards

### LINKED ISSUE #584: Re-label risk score displayed in Package Control Panel
Feedback from Andy @ GSK:

> I had one other thought, given that you’re moving to a more “metric + exploration” approach.  On the main tab there’s a score under the heading “Risk”, i.e. the score auto-generated from riskmetric.  I you’re going down the hybrid route then this is only a component and not necessarily reflective of the overall assessment.  A reviewer might decide that something with a high risk score is still not high risk after reviewing the code (or vice versa).  I think it would help to clarify by labelling it something like “Automated Risk Score”.

I think this is a good idea. I'd vote to re-label from "Risk" to "Metric Risk" or "Metric Score".

### LINKED ISSUE #587: Introduce Cards on Database Tab
Related to #383 

As we think about the final product that riskassessment app produces, we should also remember that validation efforts are a complicated and labor-intensive process that involves many different personnel to get involved. Thus, the end product can't solely be the detailed reports at package-level. To appease all parties (IT, management, others), we'll also need a high level summary of everything in the database, so I'd propose the database receives a few improvements and can be exported as a report itself.

### New cards
- [x] Total Count of pkgs in DB
- [x] Count (and %s) of pkgs by decision made / not made
- [x] Count (and %s) of pkgs by decision

AARON-CLARK:
The bulk of this PR closes #587. However, I also closed #584 since it was terribly easy to update.

Solution proposed:
image


Jeff-Thompson12 (approved):
I thought long and hard to come up with any critiques, but I cannot. Everything looks great.

Jeff-Thompson12 (changes requested):
I found something to critique!

When no decision have been made, the icon for decision categories is an X.
image

Jeff-Thompson12 (approved):
LGTM!

PR #593, opened 2023-07-21
Fix busted 'explore metrics' button on Database tab

### LINKED ISSUE #592: DB tab: "Explore Metrics" buttons busted
... after PR #574 was added. Basically, instead of navigating Risk Assessment > Maint Metrics, it will need to follow Risk Assessment > Package Metrics. Should be an easy fix.

image

AARON-CLARK:
Closes #592 

image

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: riskassessment
 Title: A web app designed to interface with the `riskmetric` package
-Version: 1.0.0.9007
+Version: 1.0.0.9008
 Authors@R: c( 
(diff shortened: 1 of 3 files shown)


Jeff-Thompson12 (commented):
@AARON-CLARK Do we want it to point specifically at the Maintenance Metrics? If that is the case, the select input needs to be updated as well.

AARON-CLARK:
> @AARON-CLARK Do we want it to point specifically at the Maintenance Metrics? If that is the case, the select input needs to be updated as well.

No, "Explore Metrics" could be any flavor of metrics, not just Maint metrics.

Jeff-Thompson12 (approved):

Issue #595, opened 2023-07-26
Add border to "Package Metrics" header/ selectInput()


AARON-CLARK:
I wish the down-arrow was enough, but it's too hard to tell that this header is also a `selectInput()` used to toggle through each page. I just adding a border around it will help users realize it's something they can toggle.

![image](https://github.com/pharmaR/riskassessment/assets/17878953/b115d0a0-99bf-4e28-9fe5-c7fa6aa96d23)

Robert-Krajcik:
Thanks @AARON-CLARK and @Jeff-Thompson12 for fixing this Norman Door.

AARON-CLARK:
closed with #599

PR #599, opened 2023-07-26
Add border back in for package metric select input

Jeff-Thompson12:
Addresses #595 

image


Jeff-Thompson12:
Don't feel like a NEWS item is needed. Can increment version if desired.

AARON-CLARK (approved):
Looks good! Well, that's not true. It looks ugly... BUT better.

PR #607, opened 2023-08-03
Set CRAN repo

Jeff-Thompson12:
A number of functions in `{riskmetric}` run `getOption("repos")`, so to ensure that the correct repo is being used, it needs to be set in the application itself.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: riskassessment
 Title: A web app designed to interface with the `riskmetric` package
-Version: 1.0.0.9010
+Version: 1.0.0.9011
 Authors@R: c( 
     person("Aaron", "Clark", role = c("aut", "cre"), email = "aaron.clark@biogen.com"),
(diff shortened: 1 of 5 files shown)


Jeff-Thompson12:
@AARON-CLARK Do you want to incorporate this into the version 2.0.0 release? It's a pretty simple change and is an important improvement in my opinion.

AARON-CLARK:
> @AARON-CLARK Do you want to incorporate this into the version 2.0.0 release? It's a pretty simple change and is an important improvement in my opinion.

Yes! The thought being that there may be times when a `riskmetric` function is executed and it won't reference the `repos` arg from `pkg_ref`?

> I can change the base to your branch if you want to include it.

`dev` is fine. Or were you implying you'd switch it to `ac-prep-200`?

Jeff-Thompson12:
In my defense, I got a little confused as to where the code was going for the next release with the `ac-prep-200` branch floating around.

AARON-CLARK (approved):

PR #610, opened 2023-08-11
demo shinyapp.io v2.0.0

Jeff-Thompson12:


Jeff-Thompson12:
@AARON-CLARK I am a little conflicted about tracking the tarballs. I suppose maybe we could track a zip file of them. Is that even desirable though? I was going to attach the zip file to this message but file size limit is 25 MB. The zip file is 10x that size.

Jeff-Thompson12:
In lieu of being able to merge `master`, I have merged `dev`.

It should also be noted that `demo-shinyapps.io` was on version 0.1.1, so I built this off of a more recent demo branch.

Jeff-Thompson12:
I just added an ignore file `.rscignore`. You might need to install a more recent version to utilize it. I was able to deploy the app to my personal shinyapps.io account with the tarballs directory intact: http://jeff-thompson51.shinyapps.io/riskassessment. It does not reach the 1GB limit.

Jeff-Thompson12:
@Robert-Krajcik Can you take a look at this and make sure everything seems in place?

PR #613, opened 2023-08-14
Fix duplicate dependencies

Jeff-Thompson12:


Jeff-Thompson12:
Choosing not to update NEWS because this fix is being pushed to `master` as well

AARON-CLARK:
I don't think this PR is necessary, is it? I would think merging `master` into `dev` would suffice, no?

Issue #586, opened 2023-07-20
Enable `DT` table filtering on Database tab


AARON-CLARK:
Related to #383 

As we think about the final product that `riskassessment` app produces, we should also remember that validation efforts are a complicated and labor-intensive process that involves many different personnel to get involved. Thus, the end product can't solely be the detailed reports at package-level. To appease all parties (IT, management, others), we'll also need a high level summary of everything in the database, so I'd propose the database receives a few improvements and can be exported as a report itself.

Allow filtering (which is a built-in feature) in DT. We just have it turned "off" right now. If there are 500 packages in here, or all of CRAN, users may want to look them up quickly

AARON-CLARK:
closed via #619

PR #619, opened 2023-08-15
Add filter for database overview

Eduardodudu:
# Task

Feature task on the option to upload all packages #586 

# This PR adds

1. Adds a filter on top of the table for database overview

image


AARON-CLARK:
Sorry @Eduardodudu, I saw this PR marked as "Draft" so I thought it was still under development. I see now that you requested my review several months ago. I pushed a couple commits to update your branch. We'll see how the CI checks fair.

AARON-CLARK:
@Eduardodudu, I also pushed a few commits to change the aesthetics of the filters. Previously, they were pretty large, using bootstrap styles. I converted to a more "plain" / minimalist look that takes up less visual real estate:

image

In addition, I'm wondering if you could explore one more thing. Right now, we've been converting everything to character so that we can manipulate how blank cells look. AKA, we've replaced `NA`s and empty chars ("") with "-". As a result, this holds us back from leveraging the cool filtering widgets that come by default with `DT`'s filter, as seen below. There are range sliders for dates and numeric values, which would come in handy. Also, factor variables get a handy dropdown menu of values to choose from, which would be useful for `decision` and `decision_by` columns. I've pushed a commit to restore these data types. This also allows us to sort columns appropriately! However, it looks like it still doesn't work unless we remove the `formattable::as.datatable(formattable::formattable())` code, which adds a lot value to our `score`, `decision`, and `explore metrics` columns. 

image

Wondering if we can somehow maintain the `{formattable}` code AND still leverage the handy filtering widgets?

AARON-CLARK:
Hey @Jeff-Thompson12, I'm getting pulled to another project for the next few days. Wondering if you can help resolve these broken tests before the new release?

Robert-Krajcik:
Hi @AARON-CLARK and @Jeff-Thompson12 
The basic filtering works.  For me, everything to the right of `score` is greyed out, probably because I haven't populated it.
Not sure filtering on `explore metrics` makes sense, but I assume it's always greyed out.

Just wondering if we could do something with the score, like allowing a range of values, or even "<" or ">"?
image

AARON-CLARK:
> ... everything to the right of `score` is greyed out, probably because I haven't populated it.

Whenever a column is populated with the a constant value, `DT` is smart enough to know that there is no variability and filtering is pointless.

> Not sure filtering on `explore metrics` makes sense, but I assume it's always greyed out.

It is, I added a little code to make sure the last column's filtering widget was disabled.

> Just wondering if we could do something with the score, like allowing a range of values, or even "<" or ">"?

Right now, `DT` does offer a range slider for numeric variables, but it's not showing up because we're using `formattable::as.datatable()` instead of `DT::datatable()`. Creating a workaround for this is going to make some extra work, so for this PR, we are going to leave it as-is.

Thanks @Robert-Krajcik! If everything else looks okay, feel free to merge.

Robert-Krajcik (approved):
Looks good to me!

Eduardodudu:
Hey @AARON-CLARK!

Sorry for the late response, I was on vacation

I see that the task is already closed.

Thanks for the return @AARON-CLARK !

PR #620, opened 2023-08-15
Re-initialize development version

Jeff-Thompson12:


Jeff-Thompson12:
May want to wait for GHAs to finish to be safe but should be good to go

Jeff-Thompson12:
@AARON-CLARK You are being very cautious with this PR.

AARON-CLARK:
You can never be too careful... pretty rigorous testing going on.

AARON-CLARK (approved):
👍🏼

Issue #621, opened 2023-08-15
Comm usage card "Package Downloads" shows as non-metric card


Robert-Krajcik:
Need to update `initialize_metric_table.sql` and possibly `test-reportPreview` test

Robert-Krajcik:
Need to update `initialize_metric_table.sql` and possibly `test-reportPreview` test

Robert-Krajcik:
Resolved with merge of PR#622

PR #622, opened 2023-08-15
Rk 621 comm metric fix

Robert-Krajcik:
Addresses issue #621

Looks like the `initialize_metric_table.sql` had already been updated.
The "Package Downloads" card is no longer flagged as non-metric

But the metric table in `upload_format.database` needs updating.

image


NEWS.md
```diff
@@ -14,7 +14,8 @@
 * Identify non-riskmetric cards (#573)
 
 ### Squashed Bugs
-* Fixed busted button introduced with #547 (#592)
+* Fixed busted button introduced with #547 
+* Fixed tests showing "Package Downloads" community usage card as non-metric
```
AARON-CLARK, line 18: You'll have to remove this from the v2.0.0 release. It should go up above, near the **development version** header.

NEWS.md
```diff
@@ -14,7 +14,8 @@
 * Identify non-riskmetric cards (#573)
 
 ### Squashed Bugs
-* Fixed busted button introduced with #547 (#592)
+* Fixed busted button introduced with #547
```
AARON-CLARK, line 17: I think we should keep the habit of tracking PRs in the news file. Could you leave the (# 592)?
```suggestion
* Fixed busted button introduced with #547 (#592)
```

AARON-CLARK (changes requested):
Looks good Robert. I just made a few comments below. Could you also increment the dev version? Thanks!

NEWS.md
```diff
@@ -1,4 +1,6 @@
 # riskassessment (development version)
+* Fixed busted button introduced with #547 (#592)
```
AARON-CLARK, line 2: Hi @Robert-Krajcik, this bullet point shouldn't be moved, correct? It belongs under "Squashed Bugs" for the `v2.0.0` release.

AARON-CLARK (changes requested):

NEWS.md
Robert-Krajcik: OK, moved it back

AARON-CLARK (approved):
👍🏼 Looks great!

Issue #626, opened 2023-08-17
Dependency "Upload All" follow up


AARON-CLARK:
Hi @Eduardodudu,

I just got around to testing #616 on `dev` and noticed that the new button exists even when the dependency table isn't rendered. Sometimes `riskmetric` errors out (due to a [known bug](https://github.com/pharmaR/riskmetric/issues/306)). Here is an example for the pkg `IDEAFilter`, with the following messaging in the console:
![image](https://github.com/pharmaR/riskassessment/assets/17878953/be24e7f6-10f0-4ef6-a295-827e99147ed0)

As such, Robert made sure the dependency table wouldn't display in these cases. I think it's appropriate that the "Upload All" button should also be hidden. In this instance, I clicked on the button and it crashed the application:
![image](https://github.com/pharmaR/riskassessment/assets/17878953/4e570aec-5117-431f-84aa-24a28d64493f)

AARON-CLARK:
closed with #628

PR #616, opened 2023-08-14
Need an "Upload All Dependencies" button #600

Eduardodudu:
# Task

Feature task on the option to upload all packages #600 

# This PR adds:

1. A structure for uploading all packages under package dependencies
2. It only updates packages that are not already uploaded
3. Reuses the component of uploading single packages
4. Added some lint for the code

image
image
image
image

PR #628, opened 2023-08-17
Dependency "Upload All" follow up#626

Eduardodudu:
# Task

Feature task on the option to hide upload all packages when there is no package to be shown in table #626 

# This PR adds

1. Checks if there are packages to be rendered on the table and only shows upload all option when it does

image

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: riskassessment
 Title: A web app designed to interface with the `riskmetric` package
-Version: 2.0.0.9001
+Version: 2.0.0.9002
 Authors@R: c( 
     person("Aaron", "Clark", role = c("aut", "cre"), email = "aaron.clark@biogen.com"),
     person("Robert", "Krajcik", role = "aut", email = "robert.krajcik@biogen.com"),
--- a/R/mod_packageDependencies.R
+++ b/R/mod_packageDependencies.R
@@ -54,6 +54,7 @@ packageDependenciesServer <- function(id, selected_pkg, user, changes, parent) {
     depends <- reactiveVal(value = NULL)
     revdeps <- reactiveVal(value = NULL)
     rev_pkg <- reactiveVal(value = NULL)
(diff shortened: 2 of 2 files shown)


AARON-CLARK (changes requested):
Hi @Eduardodudu, I know I didn't specify this request in the #626, but I think it'd also be valuable to have the "Upload All" button be hidden when there are no packages to upload. To me, it makes sense to tackle this change in this PR since it's highly related to checking for rows in the table and you already wrote the logic in #616. Specifically, your authored [this logic](https://github.com/pharmaR/riskassessment/blob/744be35f12da3af1423c339a6b19aa4d8f02fec3/R/mod_packageDependencies.R#L338-L343) which will come in handy.

https://github.com/pharmaR/riskassessment/blob/744be35f12da3af1423c339a6b19aa4d8f02fec3/R/mod_packageDependencies.R#L338-L343

Here is a use case. I already have the `zoo` package uploaded, so I don't need the option to "Upload All":

image

Eduardodudu:
> Hi @Eduardodudu, I know I didn't specify this request in the #626, but I think it'd also be valuable to have the "Upload All" button be hidden when there are no packages to upload too (see use case below). To me, it makes sense to tackle this change in this PR since it's highly related to checking for rows in the table and you already wrote the logic in #616. Specifically, your authored [this logic](https://github.com/pharmaR/riskassessment/blob/744be35f12da3af1423c339a6b19aa4d8f02fec3/R/mod_packageDependencies.R#L338-L343) which is very handy:
> 
> https://github.com/pharmaR/riskassessment/blob/744be35f12da3af1423c339a6b19aa4d8f02fec3/R/mod_packageDependencies.R#L338-L343
> 
> Here is a use case. I already have the `zoo` package uploaded, so I don't need the option to "Upload All":
> 
> image

Hi @AARON-CLARK!

I actually enjoyed a lot this idea!

Right now the app will only show "Upload All" Icon when there is actually a package to be uploaded.

PR #629, opened 2023-08-17
Standardize testing database initialization

Jeff-Thompson12:


Jeff-Thompson12:
This PR does two main things:

1. Replaces calls to `Sys.Date()` and `Sys.time()` to `get_Date()` and `get_time()` respectively. This allows us to dictate the date and time for testing purposes.
2. Creates a script in the data-raw folder that can be run for each PR. This will update the testing databases based on the branches code.

This should make the process for updating those databases more stable and easier. It also makes sure that we are running tests on the current iteration of the database structure.

AARON-CLARK:
> 2\. Creates a script in the data-raw folder that can be run for each PR. This will update the testing databases based on the branches code.

So, you say it "can be run" for each PR, but the group may need more guidance on when it should be run, explicitly. Is it just anytime the underlying database structure/schema and/or tests change? If so, could you add specific instructions to the PR checklist wiki?

Jeff-Thompson12:
Sure. I can update the wiki after merge. It is designed were running it each time should not harm, but you are right, it should be run whenever the database schema is being altered in a PR.

AARON-CLARK:
> 2\. Creates a script in the data-raw folder

Is the `data-raw` folder the best spot for this script? Just thinking out loud, but the that folder generally exists to create `.rds` files in the `data` folder, though we don't have to limit it as it's only use. I'm not against it living there because the new script does in fact create data... a db refresh in `testdata/`. Could another location be in the `dev/` folder since we expect developers to run `run_dev.R` with each PR? This new script kind of fits that bill (in scenarios where the db has changed at least). Open to hear your comments on the topic...

Jeff-Thompson12:
I don't suppose I have strong feelings for it. I always thought of the `data-raw` folder as a place to store scripts for package data and not limited to only `rds` files.

Issue #618, opened 2023-08-15
Summarize pkg dependencies


AARON-CLARK:
... could you add a summary by the "Package Dependency" count that gives a break down of how many of the dependencies fall into type "Depends", "Imports", and "Suggests" categories? Also, create a separate card that displays Base R vs "non-base R" counts. This is helpful since most all orgs consider base R pkgs trustworthy. If you want, you could mimic the database view cards [here ](https://github.com/pharmaR/riskassessment/blob/51a68005a916c77f0a95319783fd514bdaf171a9/R/utils.R#L429-L513) so they look pretty? There could be two cards: an overall count (which already exists below), and a summary card for type and base R status.

![image](https://github.com/pharmaR/riskassessment/assets/17878953/55cfe224-e0a9-4c77-a9dd-608da044cde8)

![image](https://github.com/pharmaR/riskassessment/assets/17878953/cbcb2d90-97e5-4739-b268-db16ec794b04)

_Originally posted by @AARON-CLARK in https://github.com/pharmaR/riskassessment/pull/480#pullrequestreview-1547551774_

AARON-CLARK:
closed with #636

PR #631, opened 2023-08-18
Rk 624 add suggests

Robert-Krajcik:


inst/sql_queries/initialize_metric_table.sql
```diff
@@ -14,4 +14,5 @@ VALUES
 ('dependencies',        'Dependencies',      'Number of Package Dependencies',   0, 0, 'maintenance', 1),  
 ('reverse_dependencies','Reverse Dependencies', 'Number of Reverse Dependencies', 0, 0, 'community', 1),
 ('covr_coverage',       'Test Coverage',     '% of objects tested',     1, 0, 'maintenance', 1),
-('downloads_1yr', 'Package Downloads', 'Number of downloads in last 12 months', 0, 0, 'community', 1);
\ No newline at end of file
+('downloads_1yr', 'Package Downloads', 'Number of downloads in last 12 months', 0, 0, 'community', 1),
+('suggests', 'Suggests', 'Number of Package Suggests', 0, 0, 'dependency', 0);
```
AARON-CLARK, line 18: Note: whenever you change the database in any way, you now need to run `creating_testing_databases.R` in the `data-raw` folder thanks to #629 . But that is a pretty recent change so I wouldn't expect you to be aware of it yet. Jeff did add a new bullet point to the [PR checklist](https://github.com/pharmaR/riskassessment/wiki/Developer-Pull-Request-workflows#pr-checklist), but @Jeff-Thompson12, could you also announce this new workflow and give a quick recap of what it does on slack so everyone is aware?

AARON-CLARK: Also, I think this is the first non `riskmetric` metric we are actually storing in the database. All the other ones included in the app are calculated on the fly.

AARON-CLARK (changes requested):
Hi @Robert-Krajcik, thanks for this PR! I think it will be valuable for users to see the `Suggests` pkg lists in the app. I have a few comments:

First, the button label "Hide Suggests" is wrapping for me at 100% zoom in the browser. If I reduce my zoom to 80%, it appears on one line and becomes visible again. Also, I don't think we should adopt a button in this case since it's hard to tell when the suggests pkgs are hidden vs visible, unless you inspect the table. It's even harder to tell because the `DT` default is to just show the top 15 pkgs, and the suggests may not show up in the top rows. I sorted the table so mine show up at the top. Therefore, I'd switch over to using `shinyWidgets::materialSwitch()` to solve both problems - you'll see the text next to the toggle and it will be obvious when it's "on" vs "off".

image

Or, alternatively, you could make each type into a "select button", like we did for `tidyCDISC`. That approach could be useful if users only want to skip more than just suggests. Here is the link to the code:

https://github.com/Biogen-Inc/tidyCDISC/blob/f41ed77e1fd0df86ad95ec75dbb01efa565d00af/R/mod_dataUpload.R#L25-L31

Also looking at the image above, I wonder if people would know that the 3 suggests are not counted in the 25 dependencies. I think this will be more obvious when we move to work on #618, which will show a breakdown of each group/type.

inst/sql_queries/initialize_metric_table.sql
Robert-Krajcik: OK
added the materialSwitch in place of an action button
and ran `creating_testing_databases.R`

Robert-Krajcik:
Need to checkout a quirk in the `hide_suggests` code

AARON-CLARK (changes requested):
> Using reactiveVal toggled() to prevent materialSwitch from resetting itself to FALSE every time.

Hi @Robert-Krajcik, it looks like this isn't fixed yet. It's still resetting for me every time I change a package in the control panel, or via the --> buttons. Could you look into this & verify? Thanks!

Robert-Krajcik:
It seems to be working for me, @AARON-CLARK 
Maybe I'll build a reprex.

AARON-CLARK:
@Robert-Krajcik, I deleted the db & tried restarting and terminating R. Plus, I verified that I have all the latest commits:

image

And this is what I'm seeing:

image

image

Robert-Krajcik:
Hi @AARON-CLARK 
This is ready for another review.

The only caveat I have is that packages like `{magrittr}` , `{curl}` and `{lubridate}` 
need to have the `hide_suggests` switch set to FALSE (I set `toggle(0L)` in those cases where nrow(depends()) == 0)
So if you were looking at `{forcats}` with the `hide_suggests` switch set to TRUE
and then select `{magrittr}`, the next package selected will have `hide_suggests` set to FALSE

Getting it to remember to previous `toggle()` value might take some time, but I think this is pretty good as is.

AARON-CLARK:
> The only caveat I have is that packages like `{magrittr}` , `{curl}` and `{lubridate}`
> need to have the `hide_suggests` switch set to FALSE (I set `toggle(0L)` in those cases where nrow(depends()) == 0)

Can you **not** manually override the `hide_suggests` toggle to zero for `{magrittr}` , `{curl}` and `{lubridate}` (or other pkgs where nrow(depends()) == 0)? In other words, keep things AS-IS by by displaying no table?

image

Robert-Krajcik:
I **could** do that, but that would hide the suggests packages, which I think we would want to show.
`{magrittr}` has zero dependencies, **but** it has five suggests.
image

AARON-CLARK:
> I **could** do that, but that would hide the suggests packages

I would do it! If the user is intentionally hiding the Suggests pkgs and there are no other pkg dependencies, the table should be empty. Remember, we'll be working on #618 next, so that will show a breakdown of the pkgs by TYPE. So, once that PR is developed, even though the table is empty, the summary cards at the top will display "0 Dependencies" but "5 Suggests"... so the user will know what's going on.

Robert-Krajcik:
And if the user is **not** intentionally hiding the Suggests pkgs

AARON-CLARK:
> And if the user is **not** intentionally hiding the Suggests pkgs

Yes, if they are not choosing to hide Suggests, we should show Suggests.

Robert-Krajcik:
If the user chooses to select hide_suggests, and there is data, I am showing "no data available in table"
as there really is data there; but they chose not to view it.
image

AARON-CLARK (approved):
💪🏼

Issue #483, opened 2023-04-10
Decision automation based on any individual metric


AARON-CLARK:
See comment for context: https://github.com/pharmaR/case_studies/issues/3#issuecomment-1118566145

Automatic decisions on pkg risk score make sense, but according to use cases from several pharma companies, often times these have to reject a package based on failure to meet a specific metric threshold. In the example above, Roche had to reject a few CRAN pkgs from their GxP environments because they didn't meet their thresholds for test coverage or even community usage (downloads per year). As such, we the `riskassessment` app should have a way (in the config file and via the decision automation section) to set these pkg exclusion thresholds for any metric.

Some more thoughts:
- Would need new column on pkg upload summary table when an decisions is made automatically, citing the source for the decision - maybe call it "Pkg decision reason". If accepted/ not accepted due to pkg score, the displayed value could be "pkg score with in range" However, if a metric-level exclusion was made, some text could be shown to display how that criteria wasn't met.
- If the metric is `NA`, perhaps the default behavior would be to not exclude right away, but also don't accept (if it meets the pkg risk score threshold).

Jeff-Thompson12:
@AARON-CLARK Design question: allowing decision based on individual metrics is basically setting up a process flow. The question to me is should decisions based on the risk score be kept as a unit? Or should we break them out by decision category?

In other words this:

1. metric1 rule
2. risk score rule
3. metric2 rule

compared to this:

1. metric1 rule
2. low risk score rule
3. high risk score rule
4. metric2 rule

If we break them out, those rules can be added anywhere in the process flow.

AARON-CLARK:
> If we break them out, those rules can be added anywhere in the process flow.

Wow, so related to #601. So, I haven't thought about it much! So, just to explore what that process could look like out loud, let's say we have the following rules...

1. code coverage > 50%, then high risk.
2. pkg risk score > .65, then high risk
3. No maintainer, then high risk
4. pkg has over 20 dependencies, then medium risk
5. pkg risk score > .42, then medium risk
6. no place to report bugs > .42, then medium risk

Automated decision for high risk should probably be processed first, right? So processing should occur from "high" to "low" risk? Which assumes there is an order / hierarchy to the decision categories? 

[...]

PR #632, opened 2023-08-21
Implement decision automation based on individual metrics

Jeff-Thompson12:
Addreses #483


Jeff-Thompson12:
A number of items still need to be addressed.

- [ ] Allow pre-configuration of the rules
- [ ] Apply more rigorous checkers on the filters being supplied by users
- [ ] Disabled submit buttons for in correct values
- [ ] Do we want to 

72DF8240-3AE3-4B53-9CC7-F4A47CF81501
add color to easily identify decisions applied by rules?

AARON-CLARK:
@Robert-Krajcik, wondering if you could review this PR first and I'll give it a final look?

Robert-Krajcik (dismissed):
I am able to create a list of conditionals it likes:
image


But when I go to load package `{MASS}`
image

Having some clear examples of how to specify functions would be helpful as well.

Jeff-Thompson12:
@Robert-Krajcik the second condition was returning NA. Issue should be fixed now. I also expanded the placeholder to "For example `~ is.na(.x)`, `function(x) is.na(x)`, or even `is.na`"

Robert-Krajcik:
I am getting used to working with these conditionals.
image

It stops after one failing condition, right?  So we need to put our most important conditions first?

Jeff-Thompson12:
That is correct. It returns after a decision has been made so ordering is important. Perhaps I should add a tooltip somewhere.

Robert-Krajcik:
You might also want to mention that the percentages, if they exist, are stored as whole numbers not proportions.

Jeff-Thompson12:
Well, it is making decisions based off of the assessment values themselves, so implementation of rules kind of expects the user to understand `{riskmetric}`.

AARON-CLARK:
Hey @Robert-Krajcik, just checking in on this PR. Are you done with your review? Is it ready for a 2nd review from me? Thanks!

AARON-CLARK (changes requested):
Really quick review since I have to peace out for the day.

    First thing I noticed is there isn't a lot of guidance of what's going on here. I see you have a hover-over info button, but I don't think that's enough. I think adding a short sentence under "Rule List" would help... maybe it's hidden by default and the user could toggle for "Details" to display? Either way, it could say, 

"First, choose an assessment to base the rule logic upon. Then write a equation/function to evaluate .x, where .x is the `{riskmetric}` [assessment value](https://pharmar.github.io/riskmetric/reference/index.html#package-assessment). Then, input which decision category should be assigned when your condition is `TRUE`. Decision rules can be dragged and sorted. The every time a new package is uploaded, the rules will execute in that order and will return when the first condition is met."

Thoughts? Then, some headers at the top would help. For example, you could populate 1 - 3 below with:
1. Assessment
2. Condition
3. Decision

Last, what is the rogue tick mark / single quote here on # 2 below?

image

I will give a full review tomorrow!

AARON-CLARK (changes requested):
@Jeff-Thompson12, this functionality is great! I tested it out a couple simple rules involving vignettes and they seem to work as expected. I tested out both formulas and function(x) to define conditions, and both worked well. The error catching seems to work nice too. I was writing one formula and had an extra `)` and it found it:

image

Couple of notes below:

## Running `riskmetric` code is hard
Attempting to create a rule that deems pkgs with Bug Closure Rate < 50% as High Risk, I setup the rule as follows:

image

I want to load `tidyCDISC`, which has a rate of 43.3% (confirmed by `riskmetric` code below) which should then be considered high risk:

```r
library(dplyr)
library(riskmetric)
ref <- pkg_ref("tidyCDISC", source = "pkg_cran_remote", repos = c("https://cran.rstudio.com"))

ref %>% assess_last_30_bugs_status()
 [1] FALSE FALSE FALSE  TRUE FALSE FALSE FALSE  TRUE FALSE FALSE FALSE FALSE  TRUE  TRUE  TRUE  TRUE
[17] FALSE FALSE FALSE  TRUE FALSE  TRUE  TRUE FALSE FALSE  TRUE  TRUE  TRUE  TRUE FALSE

ref %>% assess_last_30_bugs_status() %>% metric_score()
[1] 0.4333333
```

But alas, nothing happens from a decision automation perspective:

image

I think it may have something to do with how the assessment values are stored. Notice if I try to run the following `riskmetric` code which more closely mimics how the data is stored in the db:

```r
# using objects created above...
assessed <- ref %>%
  as_tibble() %>%
  pkg_assess()

assessed$bugs_status %>% metric_score()
Warning in metric_score.default(.) :
  Don't know how to score object of class "list_of_pkg_metric", "vctrs_list_of", "vctrs_vctr", "list". score is only intended to be used with objects inheriting class 'pkg_metric', returning default score of 0.
[1] 0
```
So, perhaps our assessment values don't contain their `pkg_metric' class anymore? What's odd is that it does return 0 though... which is technically less than .50, so shouldn't it have labelled my package as high risk anyway?

I did test out an alternative method in the app too (below). Knowing that if you `sum()` a logical, true values are 1 and false are 0, so I could calculate my own percentage. And this worked!

image

## Library question
Speaking of evaluating and running `riskmetric` code, what other packages do users have available to use in this sandbox area? All the pkgs we import? Everything in our `renv.lock` file?

## Align input boxes
Not a huge deal, but longer assessment values wrap down and the blue input boxes don't line up across the row. Can you maximize the height to fill the container?

image


## User feedback
So at the end of the day, how do we help users understand how the code was evaluated? Right now, it's very black boxie.

1. If a decision isn't made, that means all of the rules ran and nothing returned `TRUE`. I'm not sure what we should display to reflect this. Doing nothing is okay for now, but what if I expected something to happen (like in the example above) but don't have a way to verify the results? Do you have any thoughts? In an ideal world, It'd be nice to show a `TRUE`/`FALSE` for each condition (rule), but that's difficult to display in the Upload Summary Table. Perhaps we could tackle it in a future PR.

2.  If a decision made, I think it'd make sense to add another column to the Upload Summary Table labelled "Decision by" which could be populated with the rule that made the decision. But then we'd also have to label our rules in the decision automation module with 1, 2, 3... or A, B, C...

image

Jeff-Thompson12:
## Running `riskmetric` code is hard
This issue is related to others in that the app is currently not returning enough information to you. For instance, what is actually keeping your code from running is that you are using `metric_score()` instead of `riskmetric::metric_score()`. We are only importing certain functions from `{riskmetric}` instead of all of them, so the code was not recognizing the function you were providing. Sidenote, I will import all of `{riskmetric}` because I think that is a reasonable expectation from a user's point of view.

## Library question
Very much related to issue one. I suppose I do need to solve the problem of how to allow user functions to work in the application.

## Align input boxes
I'm not sure if this can be done with the current construction. I will look into it.

## User feedback
This is a big question that I am still struggling with. I suppose we should at least store dummy assessment data and verify that the rules will actually run before allowing a user to add them to the database. I will think more on how to display what occurred when a package is uploaded.

Jeff-Thompson12:
@AARON-CLARK I need to investigate why the `test-coverage` GHA is failing. It is failing in the reweight view test, so I don't think it is related to this PR. Plus the `R-CMD-check` is passing.

You should at least be able to check the changes you requested.

AARON-CLARK (changes requested):
🔥 PR.

Per our discussion offline, I think it would be prudent to try an implement an optional "else" decision category in case none of the rules are met!

AARON-CLARK:
@Jeff-Thompson12, is this ready for another look?

Jeff-Thompson12:
@AARON-CLARK It should be now.

AARON-CLARK (changes requested):
Hi @Jeff-Thompson12, sorry to request changes. I know this one has been out there for a while!

### Confirm `ELSE`
I just created a generic rule, with an else of "No Decision". Wondering if you could show that in the confirmation modal? It shouldn't exist as RULE 2 after clicking submit, but I think just acknowledging the selection here would be useful feedback.

image

### Re-org

I think this interface is a little hard to follow visually, and could be tightened up with a few small changes. First, could you rename a few things in this section? Namely, 
- Automate Decisions by Metric Value or Risk Score --> Decision Automation
- Rules List --> Decision Rules
- Additional Details --> Details
- Metric / Score --> Metric

image

Then perhaps a sub-header above the low, med, high checkboxes that says "By Package Risk Score". Try to make sure the new header is indented to match the current position of "Rule List", and the checkboxes could be slightly indented from the new sub header.

image

Jeff-Thompson12:
@AARON-CLARK The changes you have requested have been implemented.

AARON-CLARK (approved):
🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥

PR #634, opened 2023-08-24
Add `metric_score` visual to cards

AARON-CLARK:


AARON-CLARK:
With last commit, I abandoned the `riskmetric` logo-themed gauges (for reasons previously mentioned) for a more minimalistic gauge using the meter html tag. I added a little green to the meter whenever there is a score of zero, so people know that a score of zero is a good thing!

image

image

I think this report still looks a little sterile. It'd be nice to dress it up in some way. Also, Jeff brought up a good point regarding the license assessment value: even though `riskmetric` found data for this metric, we should still give it an orange icon ❌, treating it like the other metric's who's scores are `NA`.

FYI, changes on this branch have not been applied to the downloadable reports yet.

AARON-CLARK:
@Jeff-Thompson12, what do you think about the visual above? I was thinking that maybe we could sneak a check mark ✔️ somewhere into the card whenever the metric score = 0, so that it pops a little more and gives users the comfort and peace that comes with getting a bunch of ✔️s.

Jeff-Thompson12:
I agree that it looks a little sterile but I really like the direction it's heading. A user can easily see which metrics are negatively (or positively) affecting the overall score. It seems like a good base to work off of.

What are your thoughts of replacing 0's with checkmarks? I'm curious if that would add or take away from the visuals. If we were a little worried we could add a footnote that ✔️ = 0. Also, if we are integrating check marks, should we integrate ❌ too if the score is 1?

AARON-CLARK:
> What are your thoughts of replacing 0's with checkmarks? I'm curious if that would add or take away from the visuals. If we were a little worried we could add a footnote that ✔️ = 0. Also, if we are integrating check marks, should we integrate ❌ too if the score is 1?

I'd prefer to incorporate both the metric score (0 or 1) with the icon (✔️ or ❌) is possible, so there is no misunderstanding possible. If the score between 0 & 1, then we'd leave the meter as-is (won't add any icons).

AARON-CLARK:
Here is the new look, now available with the most recent commits:

image

image

AARON-CLARK:
Hi @Jeff-Thompson12, see recent changes above. Wondering if you could give this PR an initial review before I expand the changes to the summary reports. Also, I'm getting some testing errors that you may be able to help with. Running the `data-raw/create_testing_databases.R` is erroring out for me...

image
image

R/utils_insert_db.R
```diff
@@ -208,9 +208,14 @@ insert_riskmetric_to_db <- function(pkg_name,
       TRUE ~ as.character(riskmetric_assess[[metric$name]][[1]][1:length(riskmetric_assess[[metric$name]])])
     )
     
+    metric_score <- case_when(
+      is.na(riskmetric_score[[metric$name]][1]) ~ "NA",
+      TRUE ~ as.character(riskmetric_score[[metric$name]][1]) # rounding
```
Jeff-Thompson12, line 213: I see that you have a comment about rounding, why didn't you go ahead and round this value? Especially since you are storing it as a character.

R/utils.R
```diff
@@ -15,6 +15,50 @@ showHelperMessage <- function(message = "Please select a package"){
...
+              if(toupper(score) %in% c("NA", "NULL")) 0 
+            else ifelse(between(round(as.numeric(score), 2), 0, .08), .08, round(as.numeric(score), 2)),
+            style = "height: 30px; width: 100%;"
+          )
+        )
+    )
+  ) %>% htmltools::renderTags() # convert tagList to HTML because we can't decode '&'
+  
+  # Decode "&amp;" and convert back to HTML
+  tag_obj["html"] %>% stringr::str_replace_all("&amp;", "&") %>% HTML
```
Jeff-Thompson12, line 60: Rendering the tags does not seem to be necessary any longer to display the special characters since you are directly using `HTML()`

AARON-CLARK: Yeah, you are right. I'll remove the `renderTags()` stuff. At first, it was a hacky way to decode the automatic conversion of `&` to `&amp;` performed by `tags$label`. I used `renderTags()` to convert to html / character, then replace the `&amp;`, then convert back to HTML... not proud of it, but it worked...

AARON-CLARK:
For my future self, here is a quick reprex to test the `metric_gauge()` stuff:

```r
library(shiny)
library(dplyr)
# score = 1

metric_gauge <- function(score) { # could add id arg here
  
  tag_obj <- tagList(
    div(style = "width: 78px; text-align:center;",
        div(
          tags$label(style = "font-size:32px", # `for` = id,
             if(toupper(score) %in% c("NA", "NULL")) "NA" else {
               # flip the label display of the score to mimic the package score...
               HTML(case_when(
                  as.numeric(score) == 0 ~ "1 &#10060;",
                  as.numeric(score) == 1 ~ "0 <span style='color:#009000;'>&#10004;</span>",
                  TRUE ~ as.character(1 - as.numeric(score))
               ))
             }
          )
        ),
        div(
          tags$meter( # id = id,
            min = 0,
            max = 1,
            optimum = 1,
            low = .3333,
            high = .6666,
            value = 
              if(toupper(score) %in% c("NA", "NULL")) 0 
              else ifelse(as.numeric(score) %in% 0:.08, .08, as.numeric(score)),
            style = "height: 30px; width: 100%;"
          )
        )
    )
  ) 
}

ui <- fluidPage(
  metric_gauge("NA"),  br(),
  metric_gauge("0"), br(),
  metric_gauge(".25"), br(),
  metric_gauge(".45"), br(),
  metric_gauge("1")
)

server <- function(input, output, session) {}

shinyApp(ui, server)
```

R/utils_insert_db.R
Jeff-Thompson12: I think it would be better to store the full number (I actually don't like that we store a rounded version of the overall package score). I'm curious as to why we would store it as a character value though. It seems like it would be easy enough to store an actual number if we don't want to round.

Jeff-Thompson12:
@AARON-CLARK I really like how you can visually see how the different metrics may be affecting overall score and how you can quickly see that at a glance. I do wonder if we can find some better way to display whenever a metric is not available (like covr_coverage). I think that the yellow X's communicate a negative, where we would want to communicate something more neutral like N/A.

R/utils_insert_db.R
AARON-CLARK: yes, I think I was having trouble storing the "NA" values, so I took a shortcut. My SQL skills are getting rusty. To do it the right way, wouldn't I have to store an numeric `NA` as `NULL` in the SQL table? And then when I pull the `NULL` back into R, I'd have re-convert it to a `NA_real_`...

AARON-CLARK:
> I think that the yellow X's communicate a negative, where we would want to communicate something more neutral like N/A.

We certainly could go that route... Then we could probably abandon any and all calls to `icon()`. Or we could just choose a different icon besides the yellow X. Another option is that we could adopt the NA meter displayed in the license card below... but it would be nice to differentiate between those two scenarios. The license assessment returns a value but an NA score vs the yellow X's have an NA assessment value and NA score.

image

AARON-CLARK:
Chose to go with 'triangle-exclamation' from fontawesome:

image

AARON-CLARK:
HTML, WORD, and PDF reports now include the metric score output as well. Please test, but here is a sneak peak of some word doc output:

image

AARON-CLARK:
@Jeff-Thompson12, this is ready for a review (again). I fixed the issue with the new cards on the Deps tab. I guess the only thing that I didn't "fix" was storing the metric scores as numeric. Do you think that's necessary? Everything is working really smooth as character and I don't want to jeopardize that, lol.

Jeff-Thompson12:
@AARON-CLARK Are PDF downloads working for you locally?

Jeff-Thompson12 (changes requested):
@AARON-CLARK I really like how this has turned out.

It looks like you have some conflicts with `dev` and I requested a test or two be made for the new function `metric_gauge()`. I also can't run the PDF reports, but that is probably a local problem for me. The other reports looked good.

AARON-CLARK:
> @AARON-CLARK Are PDF downloads working for you locally?

Yes, reports are rending to PDF for me. However, I just checked and the margins are a little squished on the right-hand side. I'll seek to fix that:

image

AARON-CLARK:
Hi @Jeff-Thompson12, I added some tests as requested and I struggled to get the URLs to wrap on the PDF Rmd, so I just opted to simplify those cells and say "yes" or "no", indicating if a URL exists. Ready for re-review!

Jeff-Thompson12 (changes requested):
@AARON-CLARK Thanks for adding the tests. Few other things popped up while I was looking things over.

This occurred for the `{MASS}` package where the score didn't round.
image

It appears that some metrics will show the warning sign when in fact they have a score of 1. This is for the `{survival}` package. The score is being reported as 1 on the DOCX and PDF reports.
image

I know this issue with the PDF reports predates your PR, but if you know of a way to fix this that would be great. Otherwise we need to open another PR.
image

AARON-CLARK:
> I know this issue with the PDF reports predates your PR, but if you know of a way to fix this that would be great. Otherwise we need to open another PR.
> image

@Jeff-Thompson12, yeah, like's push this one to another PR since this PR has been lingering for a while. I actually just reverted my changes that converted those URL fields to a simple 'yes'/'no' response for the PDF report. It looked good on my screen because the package I was reviewing had a short author name & email.

I did try some other things; namely, `knitr::kable(format = 'latex', booktabs = TRUE, escape = FALSE)` paired with `kableExtra::column_spec(3, width = "3cm")` which works as expected for the comm table. However, I couldn't get it to work for the Maint table... frustrating 😵‍💫

Jeff-Thompson12:
@AARON-CLARK "Report Bugs" has been fixed, but I noticed that "Source Control" is not correct for `{MASS}`. Perhaps we should be checking if the **score** is NA or not instead of the **value** being recorded for the assessment.

Kind of a sidenote I don't know what to make of, but evidently `riskmetric::pkg_score()` is scoring "Suggests" as well.

AARON-CLARK:
@Jeff-Thompson12, I can't get anything past you! Here is the new visual with latest commit:

image

AARON-CLARK:
> Kind of a sidenote I don't know what to make of, but evidently `riskmetric::pkg_score()` is scoring "Suggests" as well.

huh?

Jeff-Thompson12 (changes requested):
It appears that the most recent changes were not applied to the HTML report.

image

AARON-CLARK:
> It appears that the most recent changes were not applied to the HTML report.

Fixed with recent commit [b70f91c](https://github.com/pharmaR/riskassessment/pull/634/commits/b70f91ccf66d1980b3c89d570abf570574d5f4fe). Any other issues you can find?

Jeff-Thompson12 (commented):
@AARON-CLARK sorry that all these finds couldn't have happened at one time. The only other thing I found is on the DOCX and PDF reports. I believe we always had the "NA"s, but I noticed "pkg_metric_error" as a value for one of the metrics when missing.
image

AARON-CLARK:
> @AARON-CLARK sorry that all these finds couldn't have happened at one time. The only other thing I found is on the DOCX and PDF reports. I believe we always had the "NA"s, but I noticed "pkg_metric_error" as a value for one of the metrics when missing.

Yeah, I saw that too and didn't change it because I thought it was a little more informative than "NA". Although, I should probably have it display what the cards display, correct? Which is "Not Found" (for 3 metrics displayed in your image). Thoughts?

Jeff-Thompson12:
> Yeah, I saw that too and didn't change it because I thought it was a little more informative than "NA". Although, I should probably have it display what the cards display, correct? Which is "Not Found" (for 3 metrics displayed in your image). Thoughts?

It is more informative, but the others are lacking information. For it to be uniform they would need to be "pkg_metric_na". However, I personally would prefer uniformity with the cards and display "Not Found".

Also, are we sure that `{riskmetric}` applies the same scoring logic to objects of class "pkg_metric_error" and "pkg_metric_na"?

AARON-CLARK:
@Jeff-Thompson12, okey dokey. Fixed it up. Specifically:
1. converted card 'value' `NA`s to 'Not found' to mimic html card process
2. Added note to docx and pdf output regarding how scores with `NA` will not impact pkg scores

image

> Also, are we sure that {riskmetric} applies the same scoring logic to objects of class "pkg_metric_error" and "pkg_metric_na"?

Not sure.

Jeff-Thompson12 (approved):
I am content now.

Issue #585, opened 2023-07-20
Migrate from `{jsTreeR}` to `{shinyTree}`


Jeff-Thompson12:
Related to #579, there is idiosyncratic handling of HTML outputs in the `{jsTreeR}` package. I have worked with the package creator to satisfactorily fix some of them but not others, particularly the one causing the issue in #579. These non-standard behaviors will make the code containing `{jsTreeR}` functions hard to maintain going forward.

While it is remarked that `{shinyTree}` does not have as many features, I believe the benefits outweigh the cost to migrate to that package instead.

AARON-CLARK:
closed with #635

PR #635, opened 2023-08-30
Migrate to `{shinyTree}`

Jeff-Thompson12:
Addresses #585


Jeff-Thompson12:
Here's a list of the cons to migrating:

1. I had to use a more recent version of the package than the snapshot we are using from RSPM. Ironically for a similar issue I had with `{jsTreeR}`, namely that it was using `.onAttach()` instead of `.onLoad()`.
2. `shinyTree()` tries to add special font-awesome styles. I have written code to remove them [here](https://github.com/pharmaR/riskassessment/blob/fca3b4df37f77dd2384a97f0b74b8ee3540fe4c5/R/mod_pkg_explorer.R#L50-L53). There does not seem to be any negative effects from doing it, but it is a bit hacky.
3. The registered inputs are not as nice to use and helper functions had to be created to gather the information needed.

Here are the pros to migrating:

1. We can use the same processing as other modules using the `uiOutput()` and `renderUI()` pairings.
2. The behavior and syntax is consistent with other widgets.
3. I have already done the work to migrate.

Jeff-Thompson12:
I also took the time to add in a progress/loader bar and spent some time beautifying it. The animation is also applied to other instances where we run `withProgress()`.

DF649A9D-231B-4098-ABF2-A83372A1A68A

Robert-Krajcik (approved):
Looks good, Jeff.
I like the updated progress bar!

PR #636, opened 2023-08-31
summarize package dependencies 

Robert-Krajcik:


Robert-Krajcik:
Note: I am counting all the unique package names by type.
Instead of a base-R summary, I created a percentage of base-R packages.

Jeff-Thompson12 (changes requested):
@Robert-Krajcik this is looking pretty good to me. I do have one request. Is it possible for the first card to be "number of dependencies uploaded to DB"? I think that would be more appropriate for the section.

Robert-Krajcik:
The dependency count will be the sum total of the package dependencies on the second card.
image
Let me know if you want any other changes

R/mod_packageDependencies.R
```diff
@@ -97,6 +97,11 @@ packageDependenciesServer <- function(id, selected_pkg, user, changes, parent) {
       suggests(pkgref()$suggests[[1]] %>% dplyr::as_tibble())
     })
     
+    cards <- reactive({
+      req(pkgref())
+      build_dep_cards(data = dplyr::bind_rows(depends(), suggests()), loaded = loaded2_db()$name)
+    }) %>% shiny::bindCache(pkgref()) %>% shiny::bindEvent(pkgref())
```
Jeff-Thompson12, line 103: By binding the cache, the total number of dependencies doesn't update if you upload one during that same session.

Robert-Krajcik: It's working now that I removed `bindCache()` thanks.

Robert-Krajcik:
Decided to show **both** count and percent of base R packages, more consistent with the "type summary" card next to it

R/mod_packageDependencies.R
```diff
@@ -160,11 +162,13 @@ packageDependenciesServer <- function(id, selected_pkg, user, changes, parent) {
         showHelperMessage()
       } else {
         req(depends())
+        # Create metric grid card.
+        metricGridServer(id = 'metricGrid', metrics = cards)
```
Jeff-Thompson12, line 166: Just noticed that the server portion of the module is embedded in a `renderUI()` resulting in multiple calls.

Jeff-Thompson12 (changes requested):
@Robert-Krajcik I tried uploading a package without any "Suggests" or other dependencies to see how the cards would display and the app crashed. I tried uploading `{abasequence}`. Error message wasn't very informative.

image

R/mod_packageDependencies.R
Robert-Krajcik: I had incorrectly assumed we would not have a problem with suggests.
Looks like I will need to do a tryCatch() like I did with depends.

And I will take a look at moving the metricGridServer() function

Robert-Krajcik: I saw your issue in the `{riskmetric}` package.
package `{abc}` only has `Depends` and so `riskmetric::assess_dependencies()` fails

Jeff-Thompson12:
@Robert-Krajcik The app is still crashing on me when trying to upload a package with no "Suggests".

Robert-Krajcik:
All set for another review @Jeff-Thompson12

Jeff-Thompson12 (dismissed):
@Robert-Krajcik The Suggests are no longer showing up for packages where riskmetric had an error pulling in the dependencies (e.g. `{magrittr}`)
image

Also, do we want no cards to be visible for packages like `{abc}`, or do we want to display something similar to the other tabs where we tell the user "None Found" or something similar.

Robert-Krajcik:
Now it looks right
image
image

AARON-CLARK (changes requested):
Hey @Robert-Krajcik, I think the user is going to forget the total number of dependencies reported by `riskmetric` and thus have to toggle back and forth to the Maint Metrics view. Can you either add a new card that mimics the `riskmetric` card or maybe add that info to the existing "Dependencies Uploaded" card? For example, instead of saying "1 (4%)" below, you could say "1 of 25 (4%)". I prefer the later.

image

image

I also think we need to shout out that you are only counting "non-suggests" packages here. I'll let you decide how you want to do that.

Last, I don't think you need to say "Base:" here sine that's the only stat your reporting on, so its redundant. However, could you change the subtitle to read: "Percent of Packages from Base R"?

image

Robert-Krajcik:
Hi @AARON-CLARK 
Cards have been updated:
image

AARON-CLARK:
> Hi @AARON-CLARK Cards have been updated: image

I'm still struggling with this because the denominators are different for card 1 vs cards 2 & 3. I probably should have suggested this last week, but what if the denominator stays consistent with whatever is shown in the table? For example, if "Hide Suggests" is `TRUE`, then all cards use the 'non-Suggest' pkg count. When it is `FALSE`, all the percentages would be calculated including 'Suggest' pkgs in the total count. That would eliminate any confusion, I think.

@Robert-Krajcik & @Jeff-Thompson12, do you have any thoughts on that?

Robert-Krajcik:
Hi @AARON-CLARK 
I think this is fine the way it is.  The first card clearly states "non-suggests" dependencies
so that the count and percent match the card in the maintenance metrics.
The other two are based on total dependencies + suggests.
Even if you select `hide_suggests` you would want to let the user know how many hidden suggests there were.

AARON-CLARK:
I think the biggest issue with keeping it as-is, would be card 1. If we care about 'Suggest' pkgs, why are we excluding those from the upload counts? If they care about suggests, they'd want to upload all pkgs, including suggests.

AARON-CLARK:
> Hold on here a minute, @AARON-CLARK
> Didn't you just say this morning

Haha, no, that was **last week**, not this morning! And this week is totally different 🤣 Just kidding - and sorry - that was a big oversight on my part. I was hoping that we could maintain the `riskmetric` dependency count somehow (and we still can). For me, the footer on the first card is too easy to glance over and potentially miss.

So, I would either (1) use all the same denominators for cards 1 - 3, or adopt (2) adopt the changing denominator that includes/ excludes 'Suggests'. Sorry for being wishy washy.

AARON-CLARK (changes requested):
Hi @Robert-Krajcik, thanks for the changes! It's definitely making more sense now with the denominator tied to the "include suggests" toggle! However, I do have just a few minor comments:

First, if we aren't including suggests, cant we just remove that Type from the Type Summary card (the 2nd card)? It just looks cleaner and I don't want anyone to think there are zero Suggest pkgs, because for this pkg I'm viewing (`tidyCDISC`) there are 3! So it's kind of misleading.

image

Also, when the denominator is zero (for pkgs like `magrittr`), can each card just report the counts and drop the percentages? The view below is a little busy with all the NaNs. Or an alternative idea would be to just state "NA" for each card. 

image

Robert-Krajcik:
Here is what `magrittr` looks like:
image
include suggests
image
If I don't add some characters like "of 0" the text runs into the next item.

Robert-Krajcik:
This is ready for another review, @AARON-CLARK

AARON-CLARK (approved):
@Robert-Krajcik, it's looking good! I wasn't a huge fan of the "0 of 0" look, so I opted to just state "0 ( 0%)" to avoid the text wrap. This how it looks:

image

PR #637, opened 2023-08-31
Add Function Explorer

Jeff-Thompson12:
Completes the second part of #498


Robert-Krajcik:
Hey this is really neat @Jeff-Thompson12 
image

Jeff-Thompson12:
@Robert-Krajcik It is pretty cool. I wish I could take credit for it. A couple of nice enhancements could be made too like being able to jump to the next highlighted row.

AARON-CLARK (changes requested):
@Jeff-Thompson12, Looks amazing! Thanks for opening this one. I'm not providing a full review here - I just had enough time to take a 15 minute peak this afternoon:

First, this PR should add the other repo to cite both our GSK contributions in the license file, similar to how we did that before for the file browser. Just note, the file browsers original repo had it's name changed to `pharmaR/pkg_dir_explorer`.

Next, wondering if you can make the width of this `selectInput()` a little wider so that the "Function Explorer" text doesn't wrap?

image

I am a little confused by the "Always Show Files" checkbox seen below. I can see that it only seems to do anything when File Type = "Man Page" in the scenario below. Does it go away when there is only 1 files available to browse? When I had zero files to browse, the file name viewer always remained no matter what I did with the checkbox. I'd say "Always show files" should always be `TRUE` and we should probably just strip that checkbox element out of the UI.

image


Also, how do you feel about the wording "Man Page?" lol I feel like we should shoot for something slightly more mainline, like maybe "Help Docs"? What do you think? Maybe whatever Hadley calls Rd files in his pkg book?

For `IDEAFilter`, why is this string showing up in the "Exported Function" dropdown?

image

Last, I like you idea about adding a button to jump to the next (or previous) use of a highlighted line in a file. If you can make it happen, knock yourself out!

I can do another review later... or maybe I'll wait until after you iterate on some of the requests above. Thanks again! This PR is 🔥

Jeff-Thompson12:
@AARON-CLARK 

1. Explorer Select Input
There is not really a way to easily adjust the overall select input based on the options. I have done a simpler solution by changing "Source Browser" to "Package Browser" which gets things to display nicer.
2. Always Show Files
I too did not like how this functioned and had adjusted it before. I agree that it is not adding any real utility and have removed it from the UI. Along with some other changes the UI is now a little crisper.
3. Man Pages
The more correct terminology is "R Documentation" which is in the top right corner of the `.Rd` files. I have edited the labels a little in all three cases. This is now "Help Documentation".
4. Exported Function Dropdown
This is a good catch. There was a little issue with the regular expression parsing the NAMESPACE file. Should be working as expected now.
5. Jump to highlighted line
This would be a nice feature but should be a separate PR. Trying to cram it in this one would just unnecessarily slow it down.

AARON-CLARK (commented):
@Jeff-Thompson12, tell me if you think this is too knit-picky (see what I did there?... knit?):

Do you feel like the help doc don't quite looks the same. I thin it's because our `bslib` theme is making the headers all caps? How do you feel about making those headers render "asis" instead? It's just coming off really strong on my monitor, lol. Not sure if that is easily achieved. Also, the help window also includes the package name in curly braces. 

image

vs

image

Jeff-Thompson12:
@AARON-CLARK We aren't going to be able to replicate the help documentation as produced in RStudio 1-1, but it should be better now. Let me know if you notice anything else that needs to be tweaked with it.

AARON-CLARK:
Hi @Jeff-Thompson12, I was looking at `magrittr`'s source code and noticed the following functions had nothing to display. I'm wondering why this module can't look up the code because I can easily find it [on GitHub](https://github.com/tidyverse/magrittr/blob/21093d06e43649a72b865811188e77fc8ad87780/R/pipe.R#L130) (or in the File Browser). Is it because the pkg author chose to use [a different @Rdname](https://github.com/tidyverse/magrittr/blob/21093d06e43649a72b865811188e77fc8ad87780/R/pipe.R#L128)?

image

https://github.com/tidyverse/magrittr/blob/21093d06e43649a72b865811188e77fc8ad87780/R/pipe.R#L128

AARON-CLARK (commented):
Also, I started exploring the code a little deeper, trying to explore the `magrittr` issues a little more and can't seem to parse the data appropriately. Do you know what I'm doing wrong here?

```r
devtools::load_all()
utils::untar(file.path("tarballs", "magrittr_2.0.3.tar.gz"), exdir = "source")
get_parse_data(type = "source", pkgdir = "/source/magrittr")
# > A tibble: 0 × 0
```

Looks like this line of `get_parse_data()` is returning `character(0)` even though the files are indeed there...

`filenames <- list.files(dirpath, ".+\\.[R|r]$")`

Jeff-Thompson12:
> Also, I started exploring the code a little deeper, trying to explore the `magrittr` issues a little more and can't seem to parse the data appropriately. Do you know what I'm doing wrong here?
> 
> ```r
> devtools::load_all()
> utils::untar(file.path("tarballs", "magrittr_2.0.3.tar.gz"), exdir = "source")
> get_parse_data(type = "source", pkgdir = "/source/magrittr")
> # > A tibble: 0 × 0
> ```
> 
> Looks like this line of `get_parse_data()` is returning `character(0)` even though the files are indeed there...
> 
> `filenames <- list.files(dirpath, ".+\\.[R|r]$")`

Remove the leading "/" in "/source/magrittrr".

I am looking into the other issue. It's a problem with the regrex and escaping characters.

Jeff-Thompson12:
@AARON-CLARK I believe everything is working as expected now. I did have to adjust a number of the regrex's so perhaps we should have one other person give the PR a whirl to make sure I didn't break anything else. The logic was not set up to handle special types of function operators. Fortunately, `{magrittr}` contains several types, so we could easily test it.

AARON-CLARK (changes requested):
Looking good!

Except I found that the `%$%` infix isn't rendering any documentation... but [it's GitHub](https://github.com/tidyverse/magrittr/blob/21093d06e43649a72b865811188e77fc8ad87780/R/pipe.R#L258-L290) shows `roxygen2` syntax. All the other pipes are now showing docs though, so it's a step in the right direction.

image

Jeff-Thompson12:
> Looking good!
> 
> Except I found that the `%$%` infix isn't rendering any documentation... but [it's GitHub](https://github.com/tidyverse/magrittr/blob/21093d06e43649a72b865811188e77fc8ad87780/R/pipe.R#L258-L290) shows `roxygen2` syntax. All the other pipes are now showing docs though, so it's a step in the right direction.
> 
> image

The regrex was having trouble with the special character "$". I have made updates to fix the problem more generally.

AARON-CLARK (approved):
💪🏼 It's so beautiful.

PR #639, opened 2023-09-01
Reduce Dependencies

Jeff-Thompson12:


Robert-Krajcik (dismissed):
Looks like everything is still working fine.
I wasn't sure how you were going to remove `{DBI}`
So we removed two packages.  OK 

I'll let @AARON-CLARK do the honors of merging this, once you update the NEWS.md file and version

Jeff-Thompson12:
@Robert-Krajcik I had removed `{DBI}` but thought better of it. Regardless `{RSQLite}` pulls in all of `{DBI}` anyways, so it wasn't a high value drop (maybe even no value).

AARON-CLARK (approved):
👍🏼

PR #642, opened 2023-09-06
Improve uploading packages from dependency tab

Jeff-Thompson12:
This PR seeks to improve on the current process in two main ways:

1. Utilizing the `trigger_events` reactive values list
The current process overwrites the list of options for adding a package. It also relies on interactions with the UI when a more direct interaction is possible and probably preferrable.

2. Remove the dependence on the reactive `changes()`
While using `changes()` works, it also causes events to trigger multiple times and other times when unneeded. Also, `changes()` is set to update whenever a decision of overall comment is added to a package. This unnecessarily complicates the reactive graph because neither of those two events are the events desired to being tracked. The reactive is merely changing as a byproduct of the package selection changing. This PR makes the reactive graph more readable in that regard.

### DIFF
--- a/R/app_server.R
+++ b/R/app_server.R
@@ -238,7 +238,6 @@ app_server <- function(input, output, session) {
   dependencies_data <- packageDependenciesServer('packageDependencies',
                                                selected_pkg,
                                                user,
-                                               changes,
                                                parent = session,
                                                trigger_events)
   
--- a/R/app_server.R
+++ b/R/app_server.R
@@ -27,7 +27,8 @@ app_server <- function(input, output, session) {
     bindEvent(credential_config$roles, credential_config$privileges)
   trigger_events <- reactiveValues(
     reset_pkg_upload = 0,
-    reset_sidebar = 0
+    reset_sidebar = 0,
+    upload_pkgs = NULL,
   )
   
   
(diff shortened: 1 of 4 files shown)


Jeff-Thompson12:
@Robert-Krajcik I'm pinging you since you are the main architect of the package dependencies module. The main reason I think the PR is needed is because of passing the vector of packages to upload. I did not like how it was overwriting the choices in the upload packages select input. I think you will agree that the change I incorporated is easy to follow and works pretty well.

Robert-Krajcik (approved):
Thanks for this PR, Jeff.
Looks great. 
No issues loading packages from dependencies tab.

AARON-CLARK (changes requested):
When reviewing dependencies for `tidyCDISC`, I tried to load `cicerone` and the app crashed with the following error:

image

I tested the same thing on our demo app (which follows `master`), and it worked fine. I did not test on `dev`. In addition, when I loaded the app back up to try again, it shows that the cicerone package was loaded into the db, but there is no score... which tells me that perhaps some info was loaded into the db but not all, which is scary! When I click the arrow to explore `cicerone`'s pkgs, the app crashes (not surprisingly) with error message "'package' not found".

image

Robert-Krajcik:
Hi @AARON-CLARK 
This error you are getting was resolved in the yet to be approved summarize package dependencies PR

AARON-CLARK:
Okay @Robert-Krajcik, thanks for letting me know. It looks like the issue does exist on `dev` after a quick verification. Even when uploading `cicerone` from the "Upload Pkgs" tab too.

PR #645, opened 2023-09-11
build report updates

### LINKED ISSUE #526: `Build Report` updates
Make sure "elements to include" doesn't reset from pkg-to-pkg. Also, it'd be nice to include these presets in the `run_app()` function. Probably not a good use of the config file.

image

In addition, after reviewing use cases and discussion with `riskmetric` team, I think it'd make sense to include/exclude certain metrics. Maybe we brainstorm up an expandable UI to add all these metrics, since real estate is tight.

Robert-Krajcik:
closes #526 [Build Report Updates]


Robert-Krajcik:
What I have here, @AARON-CLARK 
Is a modified `mod_downloadHandler_include_server()` function
that will read a file on startup, if it exists,
and keep track of user selected choices off of `input$report_includes`
across packages storing the info in `session$userData$report_includes`
and writing it out to a file at session end (when the `onStop()` condition occurs)
I would like to review it with you when you have time.

AARON-CLARK:
@Robert-Krajcik, that sounds good. Let me know when you have time to review. 

Also, when you open a PR, can you please add the github issue it closes? For example, if you say "closes #526", when this PR merges, it should automatically close that issue. Plus, it just makes it easier for me to click and review the issue. Thanks!

R/mod_downloadHandler.R
```diff
@@ -51,7 +51,8 @@ mod_downloadHandler_include_ui <- function(id){
         NS(id, "report_includes"), label = NULL, inline = FALSE,
         choices = my_choices, selected = my_choices
       )
-    )
+    ),
+    actionButton(NS(id, "save_choices"), "Save Choices")
```
AARON-CLARK, line 55: I'm not saying I'm sold on this approach yet, but I like the "Store Preferences" language a little more than "Save Choices".
```suggestion
    actionButton(NS(id, "save_choices"), "Store Preferences")
```

AARON-CLARK:
@Robert-Krajcik, I think this is a pretty cool approach to the problem at hand. However, I think one of my biggest reservations is that it only stores user preferences. Technically, an org should be able to control these report configs since they'll have governing bodies that will need to store these reports... and they probably want them all to look the same... not including any extra info.

The good news is I think it would be possible to incorporate both org and user level approaches! That is, perhaps the report configs are added to the `inst/db-config.yml` file so that the organization can set their report requirements at a global, app-level. Then, we can make user preferences a possibility using your approach, but users should always have the option (a button, perhaps?) to "Revert to Org Preferences". What do you think?

This may actually be easier when #483 because then metrics of interest will be more established in the config file, but that doesn't mean progress can't be made now.

Jeff-Thompson12:
I don't understand how this separates the preferences for users across sessions. Just doing a simple experiment locally seems to indicate it does not. If I run the app as an admin developer (i.e. `run_app(pre_auth_user = "admin")`), I can set preferences on the Build Report tab (but not the database view tab). I can then login again and get those settings. But if I then try to login as the reviewer developer (i.e. `run_app(pre_auth_user = "reviewer")`), I get the preferences that were set by the admin user.

AARON-CLARK:
Just chatted with @Robert-Krajcik on the phone we discussed a couple items:

1. Robert try to resolve the issue Jeff pointed out above
2. Have Jeff review Robert's approach to see if there is a better alternative, like the config file & the db or using a pkg like `cookies`
3. Add some sort of quick message on the screen that says "User preferences saved" for 3 seconds or so
4. It's fine to keep this PR small to solve the issue of maintaining report config selections from pkg-to-pkg. Honestly, you could do that using a reactive. However, if you want to maintain the selections from session-to-session, you could continue down this work stream. Thus, we can incorporate org-level preferences (as discussed above) in another PR. We can also address adding metric-level report configs in a separate future PR.
5. Did I missing anything?

R/mod_downloadHandler.R
Robert-Krajcik: Updated this

Robert-Krajcik:
Hi @AARON-CLARK 
I now have the preferences saved by both user and role, in case someone signed on as both an admin and a reviewer.
(but I am guessing we would expect to have one user with one role, right?)
In order to do that, I needed to make sure we were on the "Build Report" tab first.
This now displays a modal message whenever preferences are saved.
The file is written out at onStop() -- after everything has been read in.

AARON-CLARK:
@Jeff-Thompson12, wondering if you could review this PR first and I'll give it a final look?

R/mod_downloadHandler.R
```diff
@@ -40,45 +40,101 @@ mod_downloadHandler_filetype_ui <- function(id){
...
     observe({
-      # print("#####################################")
-      # print("input$report_includes:")
-      # print(input$report_includes)
-      # shinyWidgets::updatePrettyCheckboxGroup("report_includes",)
+     req(counter() == 0L)
+     req(parent$input$tabs == "Build Report")
+     
+    # retrieve user data, if it exists.  Otherwise use rpt_choices above.
+     user_file(glue::glue("./inst/report_downloads/report_prefs_{user$name}_{user$role}.txt"))
```
Jeff-Thompson12, line 88: I do not think this file location will work as currently constructed for normal use case (i.e. `riskassessment::run_app()`). I tried running the app in a different directory and I got this error upon exiting.
image

R/mod_downloadHandler.R
```diff
@@ -40,45 +40,101 @@ mod_downloadHandler_filetype_ui <- function(id){
...
-      #                   isolate(input$report_includes), my_choices)
+    }, priority = 2)
+      
+    # save user selections to session$userData$report_includes, and notify user
+    observeEvent(input$store_prefs, {
+     
+      session$userData$report_includes <- paste(input$report_includes, collapse = ",")
+      shiny::showModal(shiny::modalDialog(title = "User preferences saved",
+                                          footer = modalButton("Dismiss"), 
+                                          easyClose = TRUE))
```
Jeff-Thompson12, line 104: I don't really like how this modal looks with a title and footer but not body. Can we add something like "Report preferences stored for user" or akin to that?
image

Jeff-Thompson12 (changes requested):
@Robert-Krajcik still need to mess around with the branch a little bit more but here is my initial review.

## Database View
Report configurations are completely gone in the database view. The download on this tab is also broken.
image

## Storing Preferences
It is not clear from a user perspective that "Store Preferences" must be clicked in session to maintain the selection. For instance, if the user does not click the button and switches packages, the preferences will get reset. I do not think this is desirable behavior.

R/mod_downloadHandler.R
```diff
@@ -40,45 +40,101 @@ mod_downloadHandler_filetype_ui <- function(id){
...
+    observeEvent(input$report_includes, {
+      shinyWidgets::updatePrettyCheckboxGroup(
+        inputId = "report_includes",
+        choices = rpt_choices,
+        selected = unlist(strsplit(session$userData$report_includes,","))
+      )
+    }, once = TRUE)
+    
+    # save session$userData$report_includes to txt file at session end
+    onStop(function() {
```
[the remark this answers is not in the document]
Robert-Krajcik: Removed `onStop()`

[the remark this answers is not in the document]
Robert-Krajcik: Report configurations are back in the database view

R/mod_downloadHandler.R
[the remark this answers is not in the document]
Robert-Krajcik: user$role has been removed

Robert-Krajcik: using system.file() to construct the file name now

R/mod_downloadHandler.R
```diff
@@ -40,45 +40,101 @@ mod_downloadHandler_filetype_ui <- function(id){
...
+    counter <- reactiveVal(value = 0L)
+    user_file <- reactiveVal(value = NULL)
+    
+    output$mod_downloadHandler_incl_output <- renderUI({
+      div(
+        strong(p("Elements to include:")),
+        div(align = 'left', class = 'twocol', style = 'margin-top: 0px;',
+            shinyWidgets::prettyCheckboxGroup(
+              ns("report_includes"), label = NULL, inline = FALSE,
+              choices = rpt_choices, selected = rpt_choices
```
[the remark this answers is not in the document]
Robert-Krajcik: Sure we can do that

R/mod_downloadHandler.R
```diff
@@ -40,45 +40,101 @@ mod_downloadHandler_filetype_ui <- function(id){
...
+#' @param user char vector containing name and role
+#' @param parent the parent (calling module) session information
+#' 
+#' @importFrom shiny showModal modalDialog
+#' @importFrom glue glue
+#' @importFrom shinyWidgets prettyCheckboxGroup updatePrettyCheckboxGroup
 #'
 #' @noRd 
-mod_downloadHandler_include_server <- function(id) {
+mod_downloadHandler_include_server <- function(id, pkg_name, user, parent) {
```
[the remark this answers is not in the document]
Robert-Krajcik: removed parent parameter

R/mod_downloadHandler.R
```diff
@@ -40,45 +40,101 @@ mod_downloadHandler_filetype_ui <- function(id){
...
+      )
+    })
+    
     observe({
-      # print("#####################################")
-      # print("input$report_includes:")
-      # print(input$report_includes)
-      # shinyWidgets::updatePrettyCheckboxGroup("report_includes",)
+     req(counter() == 0L)
+     req(parent$input$tabs == "Build Report")
```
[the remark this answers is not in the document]
Robert-Krajcik: done

R/mod_downloadHandler.R
Robert-Krajcik: done

R/mod_downloadHandler.R
```diff
@@ -40,45 +40,101 @@ mod_downloadHandler_filetype_ui <- function(id){
...
+     
+    # retrieve user data, if it exists.  Otherwise use rpt_choices above.
+     user_file(glue::glue("./inst/report_downloads/report_prefs_{user$name}_{user$role}.txt"))
+     if (file.exists(user_file())) {
+       session$userData$report_includes <- readLines(user_file())
+     } else {
+       session$userData$report_includes <- paste(rpt_choices, collapse = ",")
```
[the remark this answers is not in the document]
Robert-Krajcik: replaced with a reactiveVal

R/mod_downloadHandler.R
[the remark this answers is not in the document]
Robert-Krajcik: users changing preferences for a specific package are allowed, but will only be retained if they click store preferences.

R/mod_downloadHandler.R
Robert-Krajcik: added "Report preferences stored for user" to body

R/mod_downloadHandler.R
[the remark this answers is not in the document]
Robert-Krajcik: done

AARON-CLARK:
Hey @Jeff-Thompson12, just checking in. Is your review complete? Is it ready for review # 2 from me? Thanks!

Jeff-Thompson12:
@AARON-CLARK I haven't reviewed any changes. As far as I know Robert is still working on it. @Robert-Krajcik is this ready for another look from me?

Jeff-Thompson12:
@Robert-Krajcik I was able to get the reactivity working using the `session$userData` object. I am attaching my patch. Note that I changed how the preferences were stored (i.e. separated by "\n" instead of ","). This is so that no extra parsing needs to occur on either reading or writing the values.

[userData.txt](https://github.com/pharmaR/riskassessment/files/12738699/userData.txt)

Robert-Krajcik:
Code sections:
- observeEvent(user$name) -- retrieve user data, if it exists.  Otherwise use rpt_choices defined.  Run once each tab.
- output$mod_downloadHandler_incl_output -- update checkboxes in both tabs.
- observeEvent(input$store_prefs) -- save user selections and notify user
- observeEvent(session$userData$user_report$report_includes -- update checkboxes both tabs
- observeEvent(input$report_includes -- update session$userData$user_report$report_includes

R/app_server.R
```diff
@@ -171,6 +171,25 @@ app_server <- function(input, output, session) {
...
+    "Report Author", "Report Date", "Risk Score", "Overall Comment", 
+    "Package Summary", "Maintenance Metrics", "Maintenance Comments", 
+    "Community Usage Metrics", "Community Usage Comments", "Source Explorer Comments"
+    )
+  session$userData$user_report <- reactiveValues()
+  observe({
+    req(user$name)
+
+    # retrieve user data, if it exists.  Otherwise use rpt_choices above.
+    session$userData$user_report$user_file <- system.file("report_downloads", glue::glue("report_prefs_{user$name}.txt"), package = "riskassessment")
```
Jeff-Thompson12, line 185:
```suggestion
    session$userData$user_report$user_file <- file.path(system.file("report_downloads", package = "riskassessment"), glue::glue("report_prefs_{user$name}.txt"))
```

I am currently getting errors when trying to store preferences for the first time. The suggestion is one way to fix it.
image

R/app_server.R
```diff
@@ -171,6 +171,25 @@ app_server <- function(input, output, session) {
   }) %>%
     bindEvent(selected_pkg$decision(), selected_pkg$overall_comment_added())
   
+  
+  rpt_choices <- c(
+    "Report Author", "Report Date", "Risk Score", "Overall Comment", 
+    "Package Summary", "Maintenance Metrics", "Maintenance Comments", 
+    "Community Usage Metrics", "Community Usage Comments", "Source Explorer Comments"
+    )
```
Jeff-Thompson12, line 179: I don't think we want this value populated in `app_server.R` and `mod_downloadHandler.R` especially since it will be changing as we add more to the app. I see two possible solutions:

1. We store it as an internal data value.
2. We specify it in `app_server.R` and add another value to `session$userData$user_report`.

Personally I prefer approach one. With approach one we could potentially write tests to make sure that it is getting updated with changes in the app too.

Jeff-Thompson12 (changes requested):
@Robert-Krajcik Real quick review on my end. I noticed a few things but I think we are almost there.

Robert-Krajcik:
Thanks for the review, @Jeff-Thompson12 
I will make the updates and let you know when it's ready for another review

R/mod_downloadHandler.R
```diff
@@ -40,45 +40,73 @@ mod_downloadHandler_filetype_ui <- function(id){
...
+#' @param pkg_name the name of the package passed by mod_reportPreview
+#' @param user char vector containing name and role
+#' 
+#' @importFrom shiny showModal modalDialog
+#' @importFrom glue glue
+#' @importFrom shinyWidgets prettyCheckboxGroup updatePrettyCheckboxGroup
 #'
 #' @noRd 
-mod_downloadHandler_include_server <- function(id) {
+mod_downloadHandler_include_server <- function(id, user) {
```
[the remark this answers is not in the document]
Robert-Krajcik: Done

R/app_server.R
Robert-Krajcik: copied rpt_choices to `internal-data.R` and removed from `app_server.R` and `mod_downloadHanlder.R`

R/app_server.R
```diff
@@ -171,6 +171,19 @@ app_server <- function(input, output, session) {
   }) %>%
     bindEvent(selected_pkg$decision(), selected_pkg$overall_comment_added())
   
+  session$userData$user_report <- reactiveValues()
+  observe({
+    req(user$name)
+
+    # retrieve user data, if it exists.  Otherwise use rpt_choices above.
+    session$userData$user_report$user_file <- file.path(system.file("report_downloads", package = "riskassessment"), glue::glue("report_prefs_{user$name}.txt"))
```
Jeff-Thompson12, line 179: I was thinking more about the app storage, and I think the report specifications need to be stored in the root directory like the other data. If an organization has multiple implementations of the package (say for different projects) the current storage would cause all projects to share the same storage location which is not desirable.

AARON-CLARK:
converting to draft while changes requested.

R/app_server.R
Robert-Krajcik: This directory has been moved to the db-config.yml file

Jeff-Thompson12 (changes requested):
@Robert-Krajcik The feature appears to be working as intended. Few notes below.

tests/testthat/test-generate_comm_data.R
```diff
@@ -13,12 +13,13 @@ test_that("Versions should aggregate when more than 1 version exists per year an
...
 test_that("data should contain every possible year month combo from month of first release until present month - 1", {
-  mid_years <- (min(dat$year)+1):(max(dat$year) - 1)
+  mid_years <- seq(from = min(dat$year)+1, to = max(dat$year) - 1)
   year1_mnths <- dat$month[1]:12
   yearn_mnths <- 1:dat$month[nrow(dat)]
   all_combos <- expand.grid(year1_mnths, min(dat$year)) %>%
     union(expand.grid(1:12, mid_years))%>%
     union(expand.grid(yearn_mnths, max(dat$year)))
   colnames(all_combos) <- c("month", "year")
+  expect_equal(dat %>% distinct(month, year) %>% as.data.frame, all_combos, ignore_attr = TRUE)
```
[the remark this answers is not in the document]
Robert-Krajcik: Huh. Don't know why it was left in there.  Removed.

report_preferences/report_prefs_admin.txt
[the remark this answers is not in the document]
Robert-Krajcik: I ended up deleting this file and rebuilding it
specifying `report_preferences/` should ignore all files in the folder
and I can demonstrate by creating a file
`writeLines("Hello Jeff", "report_preferences/hello.txt")`
and seeing that it is indeed ignored.

Jeff-Thompson12 (changes requested):
@Robert-Krajcik I think all that remains is to bump the version number up.

@AARON-CLARK did you want to take another look at this PR? It is technically a new feature.

AARON-CLARK:
yep, I'll take a look. Thanks for the heads up.

AARON-CLARK (changes requested):
Hi, nice work on this PR. Three things I noticed right away are shared below. Note: I terminated my R session and deleted my database after pulling in all the recent changes.

#### (1) Preferences aren't retained pkg-to-pkg
Preferences are stored from tab-to-tab (navigating from "Build Report" to "Database" tabs), and session-to-session (save preferences, log off and log back in), but not pkg-to-pkg! Here are my preferences that I stored for pkg 1:

image

I change to pkg2 in the control panel, and it goes back to my previous selections!

image)

Am I doing something wrong here?

#### (2) Why use txt files?
I don't think the app deployer should have to learn about & specify yet another folder/file system in `inst/db_config.yml` for storing these preferences. It would seem prudent to leverage our user-based credentials db to store this information, no? There could be a table to store org-level and a table for user-level preferences. I guess we can keep it as-is for this PR, but maybe we should consider moving this over to the db later? Thoughts?

#### (3) Revamp button layout
The multiple buttons in this section are going to be confusing for users, especially given their current locations. Does `store preferences` not apply to the report format selection? Shouldn't it? Perhaps we need the `Download Report` button to be in-line with the "Report Configurations" header instead of being attached to the file format selector? Also, the store preferences could also be located there. This suggestion could be pushed to another PR (or now since it's a relatively easy UI change).

image

R/mod_reportPreview.R
```diff
@@ -62,12 +62,15 @@ reportPreviewServer <- function(id, selected_pkg, maint_metrics, com_metrics,
             br(), br(),
             
             div(id = "dwnld_rp",
-              h5("Report Configurations"),
+              fluidRow(
+                  column(3, h5("Report Configurations"),),
```
AARON-CLARK, line 66:
```suggestion
                  column(4, h5("Report Configurations"),),
```

AARON-CLARK (changes requested):
Looks a lot better Robert, and the preferences are stored from pkg to pkg now.

Could you also add a `fontawesome` icon to the "Store Preferences" button? Perhaps "floppy-disk" to show that the their selections will be saved?

image

AARON-CLARK (approved):
Looks good. We may want to change the layout of the button on the Database View tab later, but I think it's fine for now.

image

PR #647, opened 2023-09-13
Catalog remotes

AARON-CLARK:
I searched all our pkg dependencies (imports) and found only 1 that was installed using GitHub... `riskmetric`! So I updated the `DESCRIPTION` file accordingly.


Robert-Krajcik (approved):
Looks OK by me

AARON-CLARK:
> @AARON-CLARK Only comment I have is that we have `{chromote}` included in Suggests and it is pointing to the "rstudio/chromote" remote. (Although the most recent CRAN version of `{shinytest2}` and `{chromote}` resolved the issues we were having earlier.)

Good catch! I forgot to do the `Suggest` pkgs. I pushed a commit adding `chromote` into the remotes section. If we decide to update them later on, we can do that in a new PR.

AARON-CLARK:
@Jeff-Thompson12, this one is ready for re-review. I should note that as soon as I changed `chromote` over to using a remote origin in the `DESCRIPTION` file, the CICD started freaking out a little, citing chromote issues (more than once) which I feel like we haven't had happen in a while. However, after re-running failed workflows, it magically worked.

PR #655, opened 2023-10-04
update mod_databaseView -- add checkbox

### LINKED ISSUE #649: Database View: select "all" pkgs for report downloader
We already have the functionality to download multiple reports, but our pkg selection is restricted by `DT` & pagination. Ideally, we should allow users to download mass quantities of reports... maybe hundreds at a time. So, once #619 is completed, I'm thinking a good use case would be a user looking to filter the list down to pkgs that were uploaded on (or after) a certain date and that were considered "low" risk, for example. Those are the reports that the org may need to submit to their "pkg inclusion" governing body.

To start, we need two things:

1. a prominent checkbox (probably below the table, nearly all the report configs) that allow users to select all pkgs, subject to any filters introduced by #619 
2. I think we need the ability to view more than 15 reports on the page so . Do we create a dropdown with 15, 50, and 100 selections? Showing "All" does make me a little nervous. I do think the checkbox above should grab all the packages defined in # 1, despite what is shown.
3. If more pkgs are shown on the page from # 2, it doesn't make sense to have the report configs below the table, so that would actually make more sense as a (collapsable?) side panel on the left-hand side.

image

Robert-Krajcik:
Closes #649 [Databse View: select "all pkgs for report downloadhanlder]


Robert-Krajcik:
Found something searching through stack overflow
image

I was using `tableid_rows_current`

AARON-CLARK (changes requested):
@Robert-Krajcik, this is working great now!

Wondering if you wouldn't mind shifting the checkbox to below the `DT` table? That way, it will be closer to the report configurations, which is why the checkbox was needed / created. That way, users will likely associate this checkbox with the the report downloads, which is what we want!

image

AARON-CLARK (approved):
Looks great!

Issue #643, opened 2023-09-08
Add comment module under new Function Explorer


AARON-CLARK:
After the merge of #637.

AARON-CLARK:
closed with #665

PR #665, opened 2023-10-17
Add comment module to function explorer

Jeff-Thompson12:
Addresses #643


AARON-CLARK (changes requested):
Hi @Jeff-Thompson12, this is looking good, BUT I have a qualm shared below:

### PDF download spins into eternity

PDF rendering get's stuck. I know this wasn't the case a few weeks ago when I was working on adding the `metric_score` to the PDF document. I did not test on `devel` but I will be optimistic and assume it's still working there because "why shouldn't it be?"

image

Can you look into this?

Jeff-Thompson12:
@AARON-CLARK Can you tell me which preferences you had selected for the report output?

Robert-Krajcik:
Hi @Jeff-Thompson12 
You might need to merge in the latest changes in dev

AARON-CLARK:
> @AARON-CLARK Can you tell me which preferences you had selected for the report output?

@Jeff-Thompson12, all elements were included / checked.

Jeff-Thompson12:
@AARON-CLARK It appears that the usage of `kableExtra` for constructing the community usage metrics could cause the download to go into limbo. I have opted to remove it since we need a more elegant solution anyways.

AARON-CLARK (approved):
👍🏼 👍🏼

Issue #667, opened 2023-10-17
Report Preferences


Jeff-Thompson12:
The app won't let you save no preferences (i.e. if you deselect all options, you cannot save the value) even though the app pretends that it is saving the preferences.

AARON-CLARK:
closed with #668

Issue #666, opened 2023-10-17
Messy reactivity with report inclusion preferences


Jeff-Thompson12:
If you quickly deselect/select report inclusion preferences, you can cause the reactivity to get stuck in an infinite loop.

AARON-CLARK:
closed with #668

PR #668, opened 2023-10-17
Fix a few bugs with choosing report options

Jeff-Thompson12:
Addresses #666 and #667


Jeff-Thompson12:
This PR also re-envisions the usage of `trigger_events`. Instead of defining a reactive values object in the server and then passing that object to all modules, it assigns the object to `session$userData` so that all modules can access it without it being passed in the arguments.

Robert-Krajcik (changes requested):
This really looks good, @Jeff-Thompson12 
But I noticed when a user has selected the "Package Dependencies" tab
and then jumps over to "Upload Packages" and uploads a package...
and then clicks on "Explore Metrics" on the "Summary of Uploaded Packages"
it takes you to the "Package Dependencies" tab
If the user is on the "Maintenance Metrics" tab, however, it works as expected.

Robert-Krajcik:
I had left the "Explore Metrics" column in the "SUMMARY OF REMOVED PACKAGE(S)"
but removed the Action button.  If  you want to remove the column, that's fine.

Jeff-Thompson12:
@Robert-Krajcik I have incorporated fixes for the bugs you found. I guess this PR has turned mostly into a squash bugs PR. Any other bugs?

Robert-Krajcik:
No other bugs encountered, @Jeff-Thompson12 
I will give this a good run-through in a few minutes

Robert-Krajcik (approved):
Looks good to me!

PR #670, opened 2023-10-18
Create new, light color-blind friendly palette

### LINKED ISSUE #654: Update colors to something lighter
We had a color-blind reviewer take a look at the app recently, and they couldn't differentiate between the default low, medium, and high risk colors very well:

image

Another user thought the yellow and red were very dark. I edited them on the fly, and they agreed these looked better (and more in-line with the aesthetic of the app's lighter theme). So, we should consider integrating something close to these, if possible. 

image

Granted, I think the default colors are determine by the number of decision categories their are, so perhaps we need to adjust our entire color palette.

Secondly, they also felt the X's and Checkmarks on the cards were really dark. Our color-blind user was confused because they both looked green:

image

AARON-CLARK:
Seeks to close #654

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -16,6 +16,7 @@
 * Build Report Updates (#526)
 * Add "Explore Metrics" button to Uploaded Pkgs Summary tab on Upload Packages tab (#653)
 * Add comment module to the Function Explorer (#643)
+* Made color palette used through the app even more colorblind-friendly and created lighter "User Guide" logo (#654)
 
 # riskassessment 2.0.0
 


tests/testthat/test-decision_automation_utils.R
```diff
@@ -5,20 +5,20 @@ test_that("assign_decisions works", {
...
   expect_equal(
     get_colors(app_sys("testdata", "skeleton.sqlite")), 
-    c("#06B756", "#A99D04", "#A63E24"),
+    c("#9CFF94", "#F8D95D", "#FF765B"),
     ignore_attr = TRUE
     )
   expect_equal(
     get_colors(app_sys("testdata", "decision_automation_ex1.sqlite")), 
-    c("#06B756", "#81B50A", "#34EBE5", "#BE7900", "#A63E24"),
+    c("#9CFF94", "#81B50A", "#34EBE5", "#BE7900", "#A63E24"), # achere
```
AARON-CLARK, line 13: Couldn't verify

tests/testthat/test-utils_config_checkers.R
```diff
@@ -1,12 +1,12 @@
...
   )
   
   expect_equal(
-    set_colors(1:20),
-    c(`1` = "#06B756", `2` = "#2FBC06", `3` = "#67BA04", `4` = "#81B50A", 
+    set_colors(1:20), # achere
+    c(`1` = "#9CFF94", `2` = "#2FBC06", `3` = "#67BA04", `4` = "#81B50A", 
       `5` = "#96AB0A", `6` = "#A99D04", `7` = "#A99D04", `8` = "#B78D07", 
       `9` = "#BE7900", `10` = "#BE6200", `11` = "#B24F22", `12` = "#A63E24", 
       `13` = "#A63E24", `14` = "#A63E24", `15` = "#A63E24", `16` = "#A63E24",
```
AARON-CLARK, line 12: couldn't verify

AARON-CLARK (commented):
@Jeff-Thompson12, couple of comments:

### Regarding text color
The black text here is very sharp relative to the soft grey colored text used everywhere else on the page. Wondering if we could tone it down so it blends in a little?

image

At the same time, we need to avoid white text color here, and apply whatever is used above.

image

And same here:

image

### Database table error
Are you getting this?

image

A little more info:

image

Jeff-Thompson12:
Database table error should be fixed with my last commit. I will look at the other items.

Jeff-Thompson12:
I think it's the bold font. Changing the color to grey doesn't look quite right. Here's what it looks like without the bold:
image

Jeff-Thompson12:
@AARON-CLARK ready for another look

PR #671, opened 2023-10-19
Force new lines in card paragraphs

Jeff-Thompson12:
Forces the new lines to be recognized in the cards. Or makes this:
image
into this:
image


Robert-Krajcik (approved):
LGTM

PR #676, opened 2023-10-20
Fix `get_exported_functions()`

Jeff-Thompson12:


AARON-CLARK (changes requested):
Hi Jeff, thanks for putting together this PR. It looks like the `zoo`, `vcd`, and `xts` pkgs looked better, but I found a couple things to note:

### Assignment arrows showing up
For pkg `zoo`:

image

### Strange Behavior

Also, just uncovered something new. When I clicked from pkg `xts` to `zoo`, I got the following error on func explorer:

image

Console message:

image

And here's line 32...

image

Jeff-Thompson12:
> ### Assignment arrows showing up
> For pkg `zoo`:
> 
> image

If you look at the NAMESPACE for `zoo` you will see that those are indeed exported functions. I believe that is the way to export the behavior for items like for example `names(x) <- x`, where it assigns values to the name attribute. But nevertheless this is helpful because the app is not identifying the source code for those functions.

Jeff-Thompson12:
@AARON-CLARK 

### Strange Behavior
This should be fixed now. I have extracted the relevant parsing from the base function.

### Assignment arrows showing up
This is proving to be a little more difficult. In `{zoo}` we have some exported functions such as `coredata<-`. However, in the source code these functions are defined as `"coredata<-" <- function(){...}`, so the parsed item is not in the list but is instead interpreted as a string constant. Including all string constants can lead to some undesirable matching. Plus we are not capturing any actual usage such as `coredata(x) <- ...`.

AARON-CLARK (approved):
Okay, So I guess there's no harm in just leaving it `as-is` since it's how the pkg author intended it. Maybe we can play with it later... or maybe we won't...

image

PR #677, opened 2023-10-20
Resolve some `renv` issues in the project

Jeff-Thompson12:


AARON-CLARK (approved):
YESSSSS! Package browser works for me now.

image

PR #683, opened 2023-10-25
Fix `selectizeInput()` on `Upload Pkgs` tab

AARON-CLARK:
Ensures users can not create and select packages from CRAN that don't exist


AARON-CLARK:
Hey @Jeff-Thompson12, just thought I'd add this quick fix to v3.0.0 too. One thing I couldn't figure out yet is how to allow for auto completion when pressing `tab`. If that's not possible, we can merge without.

Jeff-Thompson12 (approved):

Jeff-Thompson12 (approved):

PR #699, opened 2023-11-14
Adjust page margins, column wrap for pdf outputs

### LINKED ISSUE #651: Get metric value column to wrap for Maint Metric table in PDF output
> I know this issue with the PDF reports predates your PR, but if you know of a way to fix this that would be great. Otherwise we need to open another PR.
> image

I did try some other things; namely, `knitr::kable(format = 'latex', booktabs = TRUE, escape = FALSE)` paired with `kableExtra::column_spec(3, width = "3cm")` which works as expected for the comm table. However, I couldn't get it to work for the Maint table... frustrating 😵‍💫. I found that anytime I changed the format away from `knitr::kable(format = 'pandoc')`, the PDF render just spin for mins and mins, never finishing. So, I left some of that code in the .Rmd file, but just commented out. I'll initiate a new issue.

_Originally posted by @AARON-CLARK in https://github.com/pharmaR/riskassessment/issues/634#issuecomment-1729488758_

Robert-Krajcik:
Closes #651


Robert-Krajcik:
I am convinced that R Markdown and tinytex were never designed to work together.
I managed to get this small .Rmd script to (eventually) run on my Cytel laptop,
TexLive was working and was able to install all the dozen or so missing *.sty files.

Here, the system would just hang, but at least I could see the message in the *.log file

The solution for me was to install the full tinytex.  No possibility of any missing *.sty files here.

> TinyTeX-2 contains the scheme-full scheme of TeX Live, which means all LaTeX packages that you could possibly install from CTAN. This is the largest bundle, and only available in the [daily release](https://github.com/rstudio/tinytex-releases/releases/daily).

`tinytex::uninstall_tinytex()`
`tinytex::install_tinytex(bundle = 'TinyTeX-2', repository = "https://ctan.math.illinois.edu/")`

```{r, eval=FALSE}
---
title: "IrisData"
output:
  pdf_document:
    latex_engine: xelatex
---
```
  
```{r setup, include=FALSE}
options(tinytex.verbose = TRUE)
knitr::opts_chunk$set(echo = TRUE)
```

```{r}
library(knitr)
library(kableExtra)
```

```{r}
knitr::kable(iris[1:3, 1:4], format = 'latex',
      caption="Some iris data") 
```

Robert-Krajcik:
One way we can "standardize" the installation of tinytex is to have everyone just download
https://yihui.org/tinytex/TinyTeX.zip 
(this is the moderately large zip file 225+ Mb that has everything we need)
and then unzip to your AppData\Roaming folder in windows.
(I have found that using `tinytex::install_tinytex(bundle = 'TinyTeX')` doesn't always work -- missing tlmgr)
The command: `tinytex::is_tinytex()` should return TRUE after that.

Jeff-Thompson12:
I think our big problem is deployment though.

Robert-Krajcik:
FYI, when I use `format = 'latex'` for the maint metrics chunk, I see
`Underfull \hbox (badness 10000) in paragraph at lines 179--180` 
in the log.  The first page of the pdf output was written, but it just stops here.

Robert-Krajcik:
Maintenance metrics page for `{survival}` package:
image
Community usage page for `{survival}` package:
image

Both using format = 'latex'

Robert-Krajcik:
Hi @AARON-CLARK 
Everything is working now  Ready for your review.

Jeff-Thompson12:
@Robert-Krajcik I got this working on my machine as well and can see the wrapping. However, in order to accomplish this I had to follow your instructions on "standardizing" the installation of TinyTex (previously kableExtra code would not work for me). I guess my main concern is how do we set up deployment environments. Is this something that we can do on Posit Connect? If not, all this work will have to be discarded when we do #702 which is high priority.

Jeff-Thompson12:
So another approach to this problem could be to use a package such as `{pagedown}` to print the HTML outputs. We would need to spend some time making those outputs print friendly, but this has the added benefit of maintaining the look and feel of our application.

Robert-Krajcik:
Currently in the windows environment, installing the tinytex distributions don't work.
(but they do work on Posit workbench!)
Current workaround is to unzip the [TinyTeX.zip](https://yihui.org/tinytex/TinyTeX.zip) (225+ MB) file found on https://github.com/rstudio/tinytex-releases
to your Appdata/Roaming folder

image

AARON-CLARK:
Hi @Robert-Krajcik, good to hear that it works in a unix / linux env. Technically, windows is not as important, even though it's our common dev env among contributors. Regarding linux install, which command did you use to gain success? I see you reference three different commands in this PR, so I'm just curious which one I should play with myself?

First
> `tinytex::uninstall_tinytex()`
> `tinytex::install_tinytex(bundle = 'TinyTeX-2', repository = "https://ctan.math.illinois.edu/")`

Second
> tinytex::install_tinytex(bundle = 'TinyTeX')

Third
> tinytex::install_prebuilt()

One thing I think this PR needs may be a small blurb added to our "Deployments" article describing how users may need this system dependency, if they don't already have it.

Last, I think before merging this PR, we should test it on our Posit Connect instance. It looks like @Jeff-Thompson12 may have cracked the code for [system installation](https://github.com/pharmaR/riskassessment/pull/717/files#diff-3522a8c5abb070598236011d60dce0eafc7f1d70261ae9a84917dafcf04d79c8), but we can work with Biogen IT to verify.

AARON-CLARK:
Okay, converting to draft. But my comments above apply for any future PRs. I assume #723 will require syncing the `tinytex` installation?

Robert-Krajcik:
I would try using `tinytex::install_prebuilt()`

Robert-Krajcik:
Redundant with #734, which was merged yesterday

Issue #700, opened 2023-11-15
Non-credentialed deployment


Jeff-Thompson12:
We need to be able to specify a deployment (probably in the configuration file) that does not use `{shinymanager}`. This is to accommodate organizations that want to use `{riskassessment}` as a means to build reports and are not concerned with the decision making and other such features.

Jeff-Thompson12:
It appears that we can achieve usage of roles when deployed on Shiny Server or Connect by making use of the objects `session$user` and `session$groups` in a non-credentialed deployment. This would require usage of a default privilege and not requiring an admin role.

aclark02-arcus:
closed with #703

PR #703, opened 2023-11-16
Allow a non-credentialed deployment

Jeff-Thompson12:
Addresses #700


AARON-CLARK:
I think this looks really good. The only thing I would advise is to test out creating roles within our Posit Connect environment  before merging. That way, we can see if they behave as we expect them to.

Nice work!

AARON-CLARK (approved):
✔️

PR #705, opened 2023-11-22
Update `{riskmetric}` and needed dependency updates

Jeff-Thompson12:
It appears that the dependency issue is still present on riskmetric but they do have a more recent version on their main branch.


Jeff-Thompson12:
No longer needed with #709

PR #716, opened 2023-12-05
Ni 597 hyperlink dependencies

### LINKED ISSUE #597: "Hyperlink" Dependency cards to the Dependencies page
After #480 is merged...

it would be useful if we could hyperlink both cards (dependencies and reverse dependencies) to open the "Dependencies" page under the Package Metrics tab. I can open that issue.

Originally posted by @AARON-CLARK on https://github.com/pharmaR/riskassessment/pull/480#pullrequestreview-1547551774

narayanan-iyer-pfizer:
Added hyperlink to dependencies card that takes user to relevant page. 

Seeks to close #597


AARON-CLARK:
Per discussion at today's dev meeting, @narayanan-iyer-pfizer to add hyperlink from "Reverse Dependencies" card (on the Community Usage page), redirecting to Dependencies page. Thank you!

Jeff-Thompson12 (changes requested):
@narayanan-iyer-pfizer Thanks for the PR!

I'm a little conflicted about the implementation. I in general feel that modules shouldn't change or edit UI elements outside of the module (although I am guilty of doing that myself). I would prefer utilizing `session$userData$trigger_events` instead of creating Shiny inputs accessible in app_server, but this is not as big of a deal to me.

### There is no/minimal visual indication that clicking the card will do something
The only thing to indicate that you can click the card is that the cursor changes. It's hard to know the best way to let the user know they can interact with it, but I think applying a hover event to the card to give it a shadow (similar to a button) and maybe a tooltip as well would help.

### The application automatically goes to the Dependencies tab when the links are initialized
The first time you go to look at the package metrics, you get redirected to the Dependencies tab without clicking on anything. We want the initial landing to be the Metrics tab.

### You can see the button when clicking on the card.
There is a visual click event when clicking on the card. This is probably related to one of the classes you applied to the link.
Snag_aa13aa4

### Scrolling to the reverse dependencies
Perhaps we should try scrolling to the actual UI object instead of setting it to `1000` for the reverse dependencies. It doesn't seem to be scrolling down enough. This could be related to the dependency table loading (sidenote: it seems to load every time you go to the tab which isn't very desirable).

NEWS.md.bak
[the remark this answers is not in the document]
narayanan-iyer-pfizer: I have updated the code per comments. Please have a look

R/mod_metricBox.R
```diff
@@ -122,11 +132,18 @@ metricBoxServer <- function(id, title, desc, value, score = "NULL",
...
-      
-      
+        ))
+     
       
+      if (title %in% c("Dependencies","Reverse Dependencies")){ 
+         html_component <- shiny::tagAppendAttributes(html_component, 
+                                                      onMouseOver="this.style['box-shadow'] = '2px 2px 2px black';
+                                                      this.style['cursor'] = 'pointer'",
+                                    onMouseOut="this.style['box-shadow'] = '2px 2px 2px white'")
```
Jeff-Thompson12, line 142:
```suggestion
                                                      onMouseOver="this.style['box-shadow'] = '2px 2px 2px var(--bs-info)';
                                                      this.style['cursor'] = 'pointer'",
                                    onMouseOut="this.style['box-shadow'] = 'none'")
```

Jeff-Thompson12: Mostly a fan of this. I don't exactly like the black shadow since the border is a different color but my implementation above doesn't really look great either because it needs to be darker.

I would suggest removing the border on the mouse out event though.

R/app_server.R
```diff
@@ -50,8 +50,19 @@ app_server <- function(input, output, session) {
...
   
+  #click event on dependencies card to change dropdown to dependencies
+  observeEvent(list(input$`dependencies-dep_click`, input$`reverse_dependencies-dep_click`),{
+   if(!is.null(c(input$`dependencies-dep_click`,
+                 input$`reverse_dependencies-dep_click`))){
+    
+     updateSelectInput(session,"metric_type",selected = "dep")
+     if(!is.null(input$`reverse_dependencies-dep_click`)){
+     runjs(' setTimeout(function() {
+                    window.scrollTo(0,document.body.scrollHeight);}, 3000); // biding time for rev-dependecies text to load
```
Jeff-Thompson12, line 62: I'm not really a fan of the timeout but I understand why you did that. The problem is really in the dependency module which is outside the scope of this PR so I think it's fine. I tried to circumnavigate it using `session$onFlushed()` but that doesn't work either for various reasons. I think this is good for now.

Jeff-Thompson12 (changes requested):
@narayanan-iyer-pfizer This is looking pretty good to me. One small comment below on the added box-shadow elements.

R/app_server.R
narayanan-iyer-pfizer: updated code per comments

Jeff-Thompson12 (approved):

Issue #702, opened 2023-11-15
Fix PDF and DOCX downloads


Jeff-Thompson12:
These downloads are only working locally and not on the deployed instances. We need users to be able to download in the PDF and DOCX formats when the application is deployed on Posit Connect, etc.

mayank-procogia:
I have tested this extensively on my local system, shinyapps and RS Connect too.

In addition to the RS Connect, this issue is also present on deployed shinyapps app at <https://rinpharma.shinyapps.io/riskassessment/>

From the log messages, it seems that the program is not able to copy the template Word and pdf R Markdown files to the temporary directory location for parameterised render, and hence it fails.

aclark02-arcus:
Since releasing `v3.1.0` on `master`, I've deployed this branch to an internal Posit Connect server and all the downloads are still not working, even though we did fix the Word & PDF issues with the new release.

For HTML & Word:
No errors in logs

For PDF:
```! LaTeX Error: File `ucharcat.sty' not found.
Warning: Error in : LaTeX failed to compile file2fbdb72cc02fec.tex. See https://yihui.org/tinytex/r/#debugging for debugging tips. See file2fbdb72cc02fec.log for more info.
```

jthompson-arcus:
This should not be a issue with the app itself. Most likely there is a LaTeX library that needs to be installed on the Posit Connect server.

aclark02-arcus:
@mayank-procogia - do you 

> Since releasing `v3.1.0` on `master`, I've deployed this branch to an internal Posit Connect server and all the downloads are still not working, even though we did fix the Word & PDF issues with the new release.
> 
> For HTML & Word: No errors in logs

Not sure why HTML & Word weren't working before. They seem to be working just fine now... and the deployment hasn't changed. So the only problem I'm still noticing is with PDF & the error message is consistent with the above. I'll make sure this is just a system dependency issue and hopefully close this out ASAP.

LDSamson:
For if it is useful: these are all the latex packages I needed to install in a clean Docker environment to ensure .PDF reports work as expected:

```
tlmgr install multirow wrapfig colortbl pdflscape tabu threeparttable \ 
              threeparttablex environ ulem makecell xcolor selnolig \ 
	      bookmark xurl etoolbox iftex
```

PR #717, opened 2023-12-05
Fix GHAs for DOCX and PDF downloads

Jeff-Thompson12:
Addresses #702


Jeff-Thompson12:
I have investigated printing the HTML reports to PDF. While some improvements can be made to make the reports nicer to print interactively, I do not think the work to get the reports to work programmatically is worth the effort (at least not in the short term).

Efforts would be better suited to determining which LaTeX packages are necessary to be able to print the DOCX and PDF versions of the report and to provide some user feedback if those packages are missing from the environment where the application is running.

Jeff-Thompson12:
I was able to get the PDF reports to work on the GHAs by installing the `texlive-xetex` package to get XeLatex.

Jeff-Thompson12:
I think this gives us a path forward to getting the downloads to work in deployments. Hopefully the steps to get the downloads to work on the GHAs  will be similar to what will be needed on the deployment servers. The downside is that users will have to get their respective admins to install the necessary dependencies.

AARON-CLARK:
To summarize a discussion with @Jeff-Thompson12 offline: 

> It's still on draft because I haven't actually fully resolved the deployment downloads.

> I still have not created guidance for setting up that environment so that the reports can download...

>... or figuring out how to determine if a report can compile. I feel like at least we should disable downloads that can't compile (if we can). Aka, provide some user feedback if those packages are missing from the environment where the application is running.

Jeff-Thompson12:
Per discussion today, I am changing the scope of this PR since the changes made benefit our workflow. I will open a new PR to tackle the download issues when the app is deployed.

Robert-Krajcik (approved):
LGTM

Issue #598, opened 2023-07-26
Put dependency assessment rules in the config file


AARON-CLARK:
After #480 is merged...

Last: I still think it'd be nice to leverage the config file to control how the these assessments get processed, as [mentioned in a previous review](https://github.com/pharmaR/riskassessment/pull/480#pullrequestreview-1530557824):

> Some users won't care about risk scores at all, so I it would be a shame if they had to wait. Perhaps we add that option as a config file field, called "auto_assess_dependencies: FALSE". False by default. Later (much later), we could have another one called "auto_upload_dependencies: FALSE" that will automatically upload first order dependencies to the database since some pharma orgs require that. For now, maybe it looks like this in the yml, by default, leaving space for future expansion:
>
> ```
> dependencies:
>   auto_assess: FALSE
> ```

However, now I'd like to expand on that and suggest adding some more detail since there are varying levels of dependencies and all are not created equal. I would think you'd want to always review a package in the DEPENDS field since those are loaded with the package of interest. imports would be up to the user, but we'd set the default to `FALSE`. Suggests probably wouldn't be assessed much.

 ```
dependencies:
  auto_assess:
      depends: TRUE
      imports: FALSE
      suggests: FALSE
```

In addition, it might be nice to set up auto assessment of individual metrics for pkg dependencies. For example, if a user just wanted to assess code coverage for dependencies, that could be a cool option. But we should probably figure out decision automation based on a single metric first.

Originally posted by @AARON-CLARK on https://github.com/pharmaR/riskassessment/pull/480#pullrequestreview-1547551774

PR #720, opened 2023-12-06
Rk 598 dep rules config

Robert-Krajcik:
Add dependency rules to the config file
closes issue #598


AARON-CLARK (changes requested):
Hi @Robert-Krajcik, this is a nice start, but you'll also have to plug in those new configurations to something on the backend for the changes to take effect. @Jeff-Thompson12 is the expert on this stuff, but start off by doing a search for `get_db_config()` to learn how you can extract the values from the config file and use them in the app's code. So, in the example below, you'll have to run `get_db_config('dependencies')` to get a list (or a nested list?) of the sub-values you've established here.

Also, you may want to start off by adding the functionality that actually performs the auto assessment of dependencies. This would include adding a checkbox on the upload pkgs tab.

Issue #721, opened 2023-12-06
Add dependency tab's output to the report!


AARON-CLARK:
This is extremely necessary! People care about dependency info, so it deserves a prominent spot in the report.

![image](https://github.com/pharmaR/riskassessment/assets/17878953/282bb7ae-18f1-41d3-8d76-535533dc1000)

aclark02-arcus:
closed with #734

PR #723, opened 2023-12-07
Add Package Dependencies to Reports

Robert-Krajcik:
Closes issue #721


AARON-CLARK:
Hi @Robert-Krajcik, I see this PR is failing r-cmd-check and test-coverage. Those need to get resolved, but I will review anyway. Stay tuned!

AARON-CLARK (changes requested):
### Output
I noticed in the `Build Report` tab and the HTML report, there is an extra column showing up called "Name". I think that could be removed since we already have the "Package" column, keeping it consistent with the `Package Dependencies` tab.

`Package Dependencies` Tab:

image

`Build Reports` Tab & HTML report:

image

The DOCX report:
This looks a little funny in word since there is a multi-line summary being displayed. Perhaps you can play with a way to clean this up? Maybe alter the word styler file... which we don't have in the repo. Do you know what I'm talking about? Happy to provide more context. Just let me know.

image


### Suggests

Also, there is no option to include or exclude "Suggests" in this table. Honestly, this is something that I would love to be controlled from the config file (related to #720), but in the app, we can give users the option to choose "with" or "without Suggests"? How about if the "Package Dependencies" report config is selected, another material switch appears (similar to the one below) that asks users to include suggests. Perhaps it maintains state from the Package Dependencies tab.... or not if that's too tricky.

image

### Score

I also feel like the score column can be a little confusing, but only because it's not well defined. Do users know: if there is no score present in the DB, there is no score? A potential solution would be to fill these cells with "Not Assessed" or "Not uploaded". I'm open to suggestions!

### whoops

Last, this is not your problem, but I just realized the meter's color hasn't been changed in the downloadable HTML report yet (see below). I'll create a separate issue for that. 
image

Robert-Krajcik:
I am going to opt for adding `Suggests` in another PR

Robert-Krajcik:
Redundant with #734 which was merged yesterday.

PR #729, opened 2023-12-15
Improve App Startup

Jeff-Thompson12:
Since the database configurations are mostly integrated in the database itself, the application does not need to run all the checkers every single time it is initialized.


Robert-Krajcik (approved):
Looks good, startup still works.

Issue #691, opened 2023-11-07
Give option to initialize database with pkg assessments outside the app


AARON-CLARK:
Will help us with "pkg_source" and internal pkgs.

As a first step, just create a solution that continues to use "pkg_cran_remote". In a later version, we can expand this to include other sources.

aclark02-arcus:
closed with #691

PR #730, opened 2023-12-15
Upload package assessments outside of app

Jeff-Thompson12:
Addresses #691

### DIFF
--- a/R/mod_decision_automation_utils.R
+++ b/R/mod_decision_automation_utils.R
@@ -13,11 +13,11 @@
 #' @importFrom rlang is_function is_formula
 #' @importFrom glue glue
 #' @importFrom loggit loggit
-assign_decisions <- function(rule_list, package) {
+assign_decisions <- function(rule_list, package, db_name = golem::get_golem_options('assessment_db_name')) {
   decision <- ""
   dec_rule <- ""
   if (any(purrr::map_lgl(rule_list, ~ !is.na(.x$metric))))
-    assessments <- get_assess_blob(package)
+    assessments <- get_assess_blob(package, db_name)
   
   for (i in seq_along(rule_list)) {
     if (decision != "") break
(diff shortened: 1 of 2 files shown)


Jeff-Thompson12:
I believe this PR does enough to justify merging. Inside the application itself there is not much of a change from a user experience point of view. I did update the shiny notification so that the details are on a separate line. Current functionality will allow an app deployer to pre-seed the database before server deployment. A test has also been included.

Robert-Krajcik (changes requested):
Hi @Jeff-Thompson12 
I switched to the branch and did a build/install.
Then tried uploading three packages:  glue, stringr, forcats
Getting an error:

image

Jeff-Thompson12:
I can't replicate the issue but I can add in a try-catch for the tarball downloading.

Robert-Krajcik:
Hi @Jeff-Thompson12 
I am not seeing the tarball downloading errors this morning.
package upload is working fine.

`upload_pkg_lst` function working

```{r, eval=FALSE}
> pkg_list <- c("Tplyr", "pkglite", "dplyr", "forcats")
> riskassessment::upload_pkg_lst(pkg_list)
No encoding supplied: defaulting to UTF-8.
No encoding supplied: defaulting to UTF-8.
No encoding supplied: defaulting to UTF-8.
No encoding supplied: defaulting to UTF-8.
No encoding supplied: defaulting to UTF-8.
No encoding supplied: defaulting to UTF-8.
# A tibble: 4 × 4
  package version status    score
  <chr>   <chr>   <chr>     <dbl>
1 Tplyr   1.1.0   new        0.37
2 pkglite 0.2.1   new        0.32
3 dplyr   1.1.4   duplicate NA   
4 forcats 1.0.0   new        0.24
```

Robert-Krajcik (dismissed):
Looks good to me!

AARON-CLARK (changes requested):
Per discussion today:

```r
# Don't export for now while these issues persist:
# Adding a vignette example, showing how to use
# Add check to make sure db is actually there
# Note: there will be no startup initializing Credential DB using QWERTY
# User should probably supply a username and password
initialize_raa(assess_db = "database.sqlite",
              cred_db =,
              configuration = # tied to env
              )

upload_pkg_lst("Tplyr")

```

AARON-CLARK:
Per discussion on 2/27, ready for another review.

AARON-CLARK:
Hey @Jeff-Thompson12! Can you make sure you open new issue(s) to tackle/ address the items identified below?

> * Don't export for now while these issues persist:
> * Adding a vignette example, showing how to use
> * Note: there will be no startup initializing Credential DB using QWERTY; User should probably supply a username and password

AARON-CLARK (changes requested):
Hi @Jeff-Thompson12, sorry it's taken me so long to re-look at this PR. In total, it's great! But one thing I noticed is that the local (external to the app) score is different than the app-derived score. So, see if you can reproduce the following:

image

Notice the score was 0.33. Then I launched the app, deleted the `tidyr` package from the db, then loaded it on the package uploads tab. Notice the new score below is 0.25.

image

Let me know what you find. Keep in mind that I'm on a new laptop and just installed hundreds of packages to my machine... but they should all fall in line with the `renv` lock file.

AARON-CLARK:
@Jeff-Thompson12, this might help... I noticed that when `tidyr` was loaded locally (before launching the app), I got the following error message on the Package Dependencies tab. However, when I loaded `tidyr` after launching the app, there was no error here.

image

Jeff-Thompson12:
@AARON-CLARK I will take a look. I will note that we moved external uploading to a different PR. So if we believe it is functioning correctly within the application, technically we don't have a problem (yet).

Jeff-Thompson12:
@AARON-CLARK I cannot duplicate this issue. Can you verify this is still a problem on your end?

Issue #711, opened 2023-12-01
Need for separate R processes


mayank-procogia:
Bulk or time-consuming operations, such as **uploading packages**, **rendering batch reports**, and **downloading** (possibly **Package Explorer** too), can cause the current R process to become blocked, rendering the app unresponsive. This can be particularly inconvenient for concurrent users attempting to use the app simultaneously. 

Utilizing separate R processes can alleviate this issue by freeing up the base R message and ensuring a more responsive app.

aclark02-arcus:
> rendering batch reports

This is completed via #731.

Still need to implement for uploading packages

PR #731, opened 2023-12-18
Run processes in background

Jeff-Thompson12:
Addresses #711


Jeff-Thompson12:
Looking for some feedback on this PR. Interested in what features you guys think should be added or what already needs to be changed.

AARON-CLARK (changes requested):
Hi @Jeff-Thompson12,

This is pretty sweet! In my first test, I ran html reports for 3 reports, then 5 reports. When downloading 3 reports, I noticed I was still able to do things in the app, which was not expected. Is that normal? 

image

When rendering 5 reports, I was able to download a new package to the db while it was running, which is amazing. However, when it finished, the app crashed and there was an error in the console:

image

Jeff-Thompson12:
@aclark02-arcus something else I forgot to mention is that since downloading a report is now a separate event from the report creation it is possible for the report downloader or another user to change outputs. This could be as low risk as adding comments but could also include actions like deleting a package from the database, which now that I mention it could crash the app.

Issue #739, opened 2024-01-17
Development branch crashing on dependency tab


Jeff-Thompson12:
Currently the development branch is crashing when the dependency metric tab is selected for a package.

AARON-CLARK:
closed with #740

PR #740, opened 2024-01-17
Fix dependency metric tab

Jeff-Thompson12:
Addresses #739


Jeff-Thompson12:
I did not update the NEWS since this bug was not introduced on the master branch.

Robert-Krajcik (approved):
Looks good to me, Jeff.

Issue #748, opened 2024-01-18
Occasional Warning message appears when using Package Dependencies Tab


Robert-Krajcik:
For some packages, a warning message appears when selecting the Package Dependencies Tab
`Warning in format(as.numeric(comm_d1[["value"]]), big.mark = " ,") :
  NAs introduced by coercion`

AARON-CLARK:
As noted in #749 - the source of this error is coming from `riskmetric`. The app handles it "fine" so if we can figure out what's happening, we should open an issue with `riskmetric`.

PR #749, opened 2024-01-18
fix warning message package dependencies tab

Robert-Krajcik:
Adresses #748 

I'm guessing this is a `riskmetric` error but for some packages (like dplyr below) no download information is available.
This produces  
`Warning in format(as.numeric(comm_d1[["value"]]), big.mark = ",") :
  NAs introduced by coercion`
image


Jeff-Thompson12:
Okay, I'm trying to get my head around this. We clearly have download information because all the other cards are populated. It seems like something must be going wrong upstream with the `cran_logs` when the package is upload. @Robert-Krajcik what does the time graph look like for you? Is that still being produced with this dataset?

AARON-CLARK:
per discussion on 2/27, should we actually hide this warning? @AARON-CLARK to review.

AARON-CLARK:
Just curious, Is there a specific package that this is happening for?

In either case, after reviewing this a little more, I don't think we want to clear out a `pkg_metric_error` message. Especially since the  app cards are already mature enough handling the error just fine (Displaying "Not Found").

I will close this PR. But if we can identify what is happening here, we should open an issue on the `riskmetric` repo!

PR #751, opened 2024-01-22
Add feature to show reverse dependencies of selected package that are currently in database

### LINKED ISSUE #746: Feature request: show which of the packages in the database has a reverse dependency on the selected package
Helpful for assessing how a package with lower-level functions is used within an organization. At the moment it is hard to find the right package when there are many reverse dependencies.

Edit: I can probably help implement this, if others also think that this would be useful.

LDSamson:
Closes #746. 

The change will look like the screenshot below in the package dependencies tab if the active package is `dplyr` and there are two packages in the database that rely on `dplyr`:

image

Note: I have added a simple test to cover the change. This might give a multiple match warning from the `dplyr` package due to the use of `dplyr::full_join()` within the (currently not tested) function `build_dep_cards`. This warning can be solved by:
1. updating `dplyr` since the multiple match warnings thrown by dplyr’s joins are refined in  [`dplyr` version 1.1.1](https://www.tidyverse.org/blog/2023/03/dplyr-1-1-1/#revisiting-multiple-matches).
2. Use `base::merge()` instead of `dplyr::full_join()`.


LDSamson:
I realized now that the PR should probably be on the development branch in this project. I can close this and open a new one on the dev branch, if you are interested. 

A bit about me, since you probably don't know me: I am the lead Shiny developer in a small/mid-size German CRO, and we want to start using the `riskassessment` application for our R package risk assessment to use within clinical trials. This would be a small but welcome additional feature.

Jeff-Thompson12:
@LDSamson you don't need to open a new PR. We can easily change the base.

I for one really like the suggestion of this new feature. However, I think it makes more sense to have the internal reverse dependencies in a table akin to what is done for the dependencies. This would allow the user to see which version they have in the database, risk scores, and a button to quickly check them out.

Would you be interested in making those changes to your PR? If not, I can pick up the development of this feature request.

LDSamson:
Yes, that makes sense, that would make it even more useful. Good to know that you like the suggestion. I can continue on it when I have time, hopefully somewhere next week.

Jeff-Thompson12:
@LDSamson I am changing the base for this PR from `dev` to `jt-746-impr_rev_dep`. (Can't currently do it because there are no differences between the branches.) Let me know if you still have time and the desire to incorporate the change above. Otherwise, I will pick up the development on this new branch.

LDSamson:
@Jeff-Thompson12 I will assess it this weekend. If after the weekend it turns out to be too much work for me at the moment, I will let you know

LDSamson:
I've made some changes, the information looks now like this:
image

I extracted some logic in two helper functions (add_buttons_to_table() and data_table_custom()) to prevent a lot of code duplication. 

Let me know what you think of it.

Jeff-Thompson12:
@LDSamson thank you for doing this. I am currently transitioning between jobs which has kept me from reviewing this sooner but I will look at it soon.

PR #753, opened 2024-01-30
Incorporate Reverse Dependency Improvements

Jeff-Thompson12:
Addresses #746 and incorporates #751


LDSamson:
@Jeff-Thompson12 I implemented suggested changes on my own fork since that's where I have write access; see #751.

Jeff-Thompson12:
@AARON-CLARK & @Robert-Krajcik I still haven't done a deep dive into the code add by this PR. I would like your input on how the output is being produced.

For example, consider the two tidyverse packages of {dplyr} and {tibble}. {tibble} is in the imports of {dplyr}, and {dplyr} is in the suggests of {tibble}. Currently the application outputs both as reverse dependencies of each other. The table currently does not include a type for reverse dependency and we could include it but I it might be confusing. Then we would have a table where package: tibble and type: imports would tibble is in the imports of {package} but the second table it would mean {package} is in the imports of tibble. Thoughts?

AARON-CLARK:
> @Jeff-Thompson12 I implemented suggested changes on my own fork since that's where I have write access; see #751.

@LDSamson I granted you write permissions! Done ✅

aclark02-arcus:
@Jeff-Thompson12, I would think adding a "type" column could make things slightly confusing, but maybe not. What if we moved the new table below the text output of "All reverse deps"? And we could make the "Reverse Deps" header larger. That may help folks separate the two tables in their minds.

Regardless if it's added or not, I did notice that `dplyr` would show up as a reverse dependency of `tibble` even when the `Include Suggests` toggle was off (or `FALSE`). I would think the `Include Suggests` toggle would apply to all output on this tab. If it's toggled on (`TRUE`), then `dplyr` should show up in the new table called "Reverse Deps available in DB". Otherwise, when `FALSE`, it should disappear. Thoughts?

Jeff-Thompson12:
@aclark02-arcus It's ready for you to take another look.

@LDSamson Sorry it's taken so long to get this PR ready to review. If some desired functionality was lost, feel free to request changes.

AARON-CLARK (changes requested):
Hi @jthompson-arcus, I'm not sure if this is happening for you, but I'm getting a blank Package Dependency tab when `tidyCDISC` is selected on the control panel. Can you please verify?

aclark02-arcus (approved):
Reviewing after recent commits, everything looks good! Thanks @LDSamson & @jthompson-arcus! Merging this one. If there is anything else missing, it could be incorporated as a new PR.

PR #759, opened 2024-03-25
Update report logo to light-theme

### LINKED ISSUE #737: App logo outdated in reports
For HTML / DOCX / PDF

image

We should seek to mimic the shiny app's logo, which has been a lighter rendition for a few months now:

image

aclark02-arcus:
Closes #737
Swaps out logo so that reports end up with the lighter version of the app's logo, instead of the older, darker themed logo.


Jeff-Thompson12 (approved):
LGTM

PR #764, opened 2024-04-22
Update R and package versions

### LINKED ISSUE #762: Upgrade Version of R
Currently using `v4.2.2`

Check what the latest version is...
Update `shiny` and other packages as needed

jthompson-arcus:
Resolves #762

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -38,7 +38,7 @@ BugReports: https://github.com/pharmaR/riskassessment/issues
 Encoding: UTF-8
 LazyData: true
 Roxygen: list(markdown = TRUE)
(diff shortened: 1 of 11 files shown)


jthompson-arcus:
> I tried to play around with all the functionality in the app as possible, and the only thing I found so far is the following error when uploading a package from the `Dependencies` tab. It happened for all the pkgs I tested, but note that it didn't really impact the app... it didn't crash and all the functionality was still there.
> 
> image

I'm having trouble replicating this. Was it a specific package? Or any package?

aclark02-arcus (approved):
This looks good to me. Game changer. Thanks for bringing the repo into 2024!

PR #767, opened 2024-04-24
Add write-manifest.yaml

jthompson-arcus:


jthompson-arcus:
@aclark02-arcus check it out. I originally opened this PR on `master` so that the GHA would trigger. If you look at the yaml you will notice that I had to pull in the most recent version of `rsconnect` for the build to work. When we update R and the PPM repo date, we should be able to pull from our snapshot.

Obviously, the first time it is run it will cause the most line changes, but check it out and let me know what you think.

aclark02-arcus:
Looks good!

jthompson-arcus:
I want to point out that if you look at 1a8bf23, you will see that despite the manifest being rebuilt, there is very few line changes for the `manifest.json` file. The changes are a result of updated a [line](4f2feca) in the file. This is showing the desired behavior or this PR, where the manifest is being built in a more standardized way.

aclark02-arcus (approved):
Outstanding!

Issue #774, opened 2024-05-03
New Package Dependency Card - red flag deps


aclark02-arcus:
Just thinking out loud, maybe we need to add some more actionable insights here, like "Number of non-low risk deps". So anytime a package dependency is not marked "low risk" (in this config's case - marked as medium or high risk), it would count that. We could assign that card the `danger` class so it looks scary. In fact, maybe we need to apply the `danger` class to the "Dependencies Uploaded" card if the percent is <100%.

Thoughts?

_Originally posted by @aclark02-arcus in https://github.com/pharmaR/riskassessment/pull/750#pullrequestreview-2038073834_

aclark02-arcus:
closed with #787

PR #744, opened 2024-01-17
Fix application crash

Jeff-Thompson12:
Addresses #741 in part. Does not add tests but seeks to fix current crashes.

PR #775, opened 2024-05-03
Clean dependency cards

aclark02-arcus:
Performing a couple of house cleaning items before working on #744

image

This PR adds the "danger" class to the icon and card value when the percent of uploadable packages is not 100%. Notice packages `cli` good examples of only 1 base R package and no other pkgs and `RColorBrewer` is good examples of no dependencies at all for testing purposes. In those cases, the class should remain "information" since it's not possible to upload any pkgs. A easy pkg to test for getting to 100% is `bayesAB` since it only requires 4 pkgs be uploaded. Notice how the percent is 80% below, but the text and icon are blue since the user can't upload the last pkg, which is a base-R package.

For some reason, these "danger" class changes are not working on the html reports though.

image

I also changed up the icon & title for the third card, which calculates the percent of deps that are base-R. I could see this card morph in the future to display all Trusted package vendors (like tidyverse, or other maintained lists).

In this PR, I may or may not add the decision column to the table of dependencies... I might wait for #774.


aclark02-arcus:
Ready for a look!

R/utils_build_cards.R
```diff
@@ -238,13 +238,17 @@ build_dep_cards <- function(data, loaded, toggled){
     succ_icon = character(),
     icon_class = character(),
     is_perc = numeric(),
-    is_url = numeric()
+    is_url = numeric(),
+    type = "information"
```
jthompson-arcus, line 242:
```suggestion
    type = character()
```

Since all the other values are empty, `dplyr::tibble` is forcing this to `character()` anyways, but I think we should have it coded as an empty value anyways.

jthompson-arcus (changes requested):
One small comment below. I also noticed that the color coding is not in the HTML report download. Don't we want to add it there too?

aclark02-arcus:
> For some reason, these "danger" class changes are not working on the html reports though.

Yeah @jthompson-arcus, I was having trouble getting this to work.

I accepted your other change.

jthompson-arcus (approved):
LGTM

PR #780, opened 2024-05-15
Highlight individual contributors more than orgs on `About` tab

### LINKED ISSUE #778: Remove pharma org logos from `About` tab
We don't necessarily have organizations participating, but we do have individuals. So stating the organizations are committed is a liability.

- Could mention orgs as plain text, with some styling (better than bulleted list... keep the grid)
- Could show pics of people, their names, and name of company

aclark02-arcus:
Fixes #778

Here's an example of what this PR accomplishes:

image

I think it looks really good, but a couple of things I'm noticing right off the bat is when the person's name is too long to fit within the width of the card, it get's left justified automatically. 

Also, the `Past Contributors` section contains cards that are noticeably taller. See below for examples. I tried toggling the height, but it didn't seem to work as expected. I also tried to make sure my images were the exact same width and height. I'm not convinced that worked either. If someone can help me figure this out, I'd be much appreciative!

image

Once polished, I'd like to write some `testthat` tests for the new functions, and I'd also like to invite current and past members to either (1) give me their permission to use their photos or (2) upload their own.


NAMESPACE
```diff
@@ -8,6 +8,7 @@ export(run_app)
 export(showComments)
 import(dplyr)
 import(riskmetric)
+import(shiny)
```
jthompson-arcus, line 11: We don't want to import all of shiny.

R/mod_aboutInfo_utils.R
```diff
@@ -0,0 +1,122 @@
...
+#'
+#' @param role character, role within the project, usually 'Workstream Lead',
+#'   'Core Contributor', or 'Contributor'
+#' @param pic character, file path to image, 1.3" in width
+#' @param site character, if they have a personal of GitHub site to reference
+#' @param name character, first and last name of the contributor
+#' @param org character, a organization that employes this individual, if
+#'   applicable
+#'
+#' @import shiny
```
jthompson-arcus, line 16: This is the offensive line that imports all of shiny.

R/mod_aboutInfo_utils.R
```diff
@@ -0,0 +1,122 @@
...
+      class = "d-flex justify-content-between",
+      role
+    ),
+    bslib::card_body(
+      class = "align-items-center",
+      gap = 0,
+      fillable = TRUE,
+      bslib::card_image(
+        border_radius = "top",
+        file = pic,
```
jthompson-arcus, line 41:
```suggestion
        file = NULL,
        src = pic,
```
Currently, you have these photos bundled into the application itself, which I think means we want to point to their source directly using the suggestion above. Why do both?

R/mod_aboutInfo_utils.R
```diff
@@ -0,0 +1,122 @@
...
+contrib_card <- function(role, pic, site, name, org){
+  bslib::card(
+    bslib::card_header(
+      class = "d-flex justify-content-between",
+      role
+    ),
+    bslib::card_body(
+      class = "align-items-center",
+      gap = 0,
+      fillable = TRUE,
```
jthompson-arcus, line 38:
```suggestion
      fillable = FALSE,
```
This will keep the photos from getting stretched out.

R/mod_aboutInfo_utils.R
```diff
@@ -0,0 +1,122 @@
...
+#' make_contrib_cards(team_info_df %>% filter(status %in% "past"))
+#'
+#' @keywords internal
+#' @family contrib_cards
+#' 
+make_contrib_cards <- function(df = team_info_df){
+  contrib_group <- purrr::pmap(df, function(role, photo_file, site, name, org, ...){
+    contrib_card(
+      role = role,
+      pic = file.path('inst/app/www/images',photo_file),
```
jthompson-arcus, line 105:
```suggestion
      pic = file.path('www/images',photo_file),
```
If we make the change above to point directly to source, we need to make this change as well.

jthompson-arcus (changes requested):
I think also if we have some standardized dimensions (or even aspect ratio) to the photos, then we will be able to more easily get the two rows of cards to have the same height. Either way that will help us with formatting.

R/mod_aboutInfo_utils.R
jthompson-arcus: We could also probably leave this be if we standardize the size/ratio of the photos.

aclark02-arcus:
Thanks for all the suggestions @jthompson-arcus, I was able to incorporate them all and add a few more changes. Below is what is rendered now. I basically realized that all our GitHub sources for pics were squares, so I decided to stay with the square theme throughout.

image

A couple of things that are still out of sorts:

- looks like we lost the center-justified-items class, but this doesn't look terrible left justified
- the target = "_blank" stuff still isn't working

R/mod_aboutInfo_utils.R
```diff
@@ -0,0 +1,122 @@
...
+      class = "align-items-center",
+      gap = 0,
+      fillable = FALSE,
+      bslib::card_image(
+        border_radius = "top",
+        file = NULL,
+        src = pic,
+        href = ifelse(is.na(site) | is.null(site), 'https://www.pharmar.org', site),
+        target = "_blank" # doesn't work as hoped
+      ),
```
jthompson-arcus, line 45:
```suggestion
        href = ifelse(is.na(site) | is.null(site), 'https://www.pharmar.org', site)
      ) |>
        tagAppendAttributes(target = "_blank", .cssSelector = "a"),
```
This fixes the target for the URLs.

dev/001_start.R
```diff
@@ -115,6 +115,7 @@ usethis::use_package("glue")
 usethis::use_dev_package("riskmetric")
 
 usethis::use_package("rstudioapi", type = "Suggests")
+usethis::use_package("png", type = "Suggests")
```
jthompson-arcus, line 118: I noticed that this is "added" to the project but not `renv.lock`. I'm assuming you used it to resize the pictures. Perhaps we should ignore the `dev/` folder with regards to `renv`.

tests/testthat/test-contrib_card_utils.R
```diff
@@ -0,0 +1,50 @@
...
+test_that("team_info_df has the 'most crucial' columns", {
+  
+  testthat::expect_true(all(c("name", "role", "site", "org", "photo_file") %in% names(team_info_df)))
+  
+})
+
+# Make sure the png's are squares
+test_that("Pngs formatted as squares", {
+  
+  if(rlang::is_installed("png")) {
```
jthompson-arcus, line 13: I see that `png` is used here. Should be added to `renv.lock`.

jthompson-arcus (changes requested):
@aclark02-arcus I think the cards are looking really good now. Looks like the package `png` needs to be added to the R environment for the test, but other than that I'm satisfied.

Want to change it from draft? Or do you have other changes you want to make?

jthompson-arcus (approved):
LGTM

PR #782, opened 2024-05-31
Fix delete package bug

### LINKED ISSUE #718: Track Package Repo
Track which repository a package was uploaded from to the application. This should be captured on a package to package basis and should be stored in the database.

jthompson-arcus:
Closes #718


aclark02-arcus (changes requested):
Works like a charm! However, I don't think you added a bullet point to the news file for your last PR (incorporating decisions in the CSV upload file). Do you kind adding a line to this PR? Thanks!

aclark02-arcus (approved):
LGTM

PR #784, opened 2024-05-31
Add `Decision` col to Pkg Dependency table

aclark02-arcus:
Seeks to close the first part of #774 with a start of adding the `Decision` column to the pkg dependency tables. Recall, there are three spots this table shows up: the pkg dep module, the report preview mod, and the reports themselves.

I think everything is kosher, and I even managed to enhance the "build report", which previously wasn't rendering the `score` and `decision` fields with color. It appears to be a little tricker to do this on the HTML reports, so I didn't tackle it in this PR.

Two pesky things I noticed in this PR is...

(1) that the menu won't disappear on the Build Report view:

image

(2) the console keeps showing this message when the pkg deps tab and build report tabs are opened:

image


aclark02-arcus:
> 2\. I don't see why decision_id is being pulled in. Am I missing something?

`decision_id` will be needed for part II of this PR. I just didn't want to mess with that query twice. To summarize, I want to build a single card that communicates two new metrics for any given pkg:

- the number of dependencies that are the highest decision category (presumably the highest risk), and...
- the number of dependencies that are not the lowest category (presumably the non-low risk category)

I think both stats would be useful to know. So I need `decision_id` to do that. Do you have any comments or critiques on that approach. I think it works and I'll add a sentence [here](https://pharmar.github.io/riskassessment/articles/Deployment.html#decisions-configuration) to coach deployers to know that they should list their categories from lowest risk to highest risk. I guess I'll add some special logic that if only two decision categories exist, than only the top statistic is needed.

jthompson-arcus:
That makes sense to me. We have so much logic in the app that assumes an ordering to the decision categories anyways. I think it's natural to assume that even if an org wants additional categories for workflow purposes, those could be housed between the reject/accept decisions.

aclark02-arcus:
@jthompson-arcus: updated!

image

jthompson-arcus (changes requested):
Couple more things:

1. `NA`s are still showing up on the PDF reports for the decision column.
2. The table on the build report/viewer tab is showing `-1` for number of entries to view. (Not sure if your PR introduced this but if you could fix it that would be great.)
image

jthompson-arcus (approved):
Console warning is related to `get_text_color()` function. It is expecting a single value but getting a list. It also shows up when viewing the database viewer. I will push a fix in a separate PR.

PR #787, opened 2024-06-04
Dep Card: Decision Summary

### LINKED ISSUE #774: New Package Dependency Card - red flag deps

aclark02-arcus:
Part II of a two part PR to close #774. It also updates that layout of the package dependencies tab (and reports) as well.

Some views of the new card:
- When no high risk exists:

    image

- When there is any high risk package, the text and icon turn red:

    image

- When no decisions exist (or no dependencies)

    image


jthompson-arcus (changes requested):
Few things:

1. I noticed that the red color when a dependency has a high decision doesn't show up on the HTML report. Is this intentional or an oversight?
2. In the PDF report I didn't get the line break for the new card.
Snag_10141cda

3. Getting this warning in the console when "Include Suggests" is selected
Snag_10157c99

jthompson-arcus (changes requested):
Can you update the NEWS?

jthompson-arcus (approved):
LGTM

PR #788, opened 2024-06-04
Fix `get_text_color()`

jthompson-arcus:


aclark02-arcus (approved):
Appears to have worked!

PR #789, opened 2024-06-06
Repair interconnectivity of report inclusions

jthompson-arcus:
Addresses a bug in the report inclusions.

To reproduce the bug on the `dev` branch do the following:

1. Select a package.
2. Go to Build Report.
3. Change the report inclusions (e.g. deselect "Report Author")
4. Go to different tab (e.g. Package Metrics)
5. Switch to a different package.
6. Return to Build Report.
7. Observe that your changes in Step 3 are not included

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -4367,7 +4367,7 @@
         "Maintainer": "Kevin Ushey <kevin@rstudio.com>",
         "Repository": "RSPM",
         "Date/Publication": "2024-02-29 01:10:07 UTC",
-        "Built": "R 4.3.0; ; 2024-03-01 11:32:02 UTC; unix"
+        "Built": "R 4.3.3; ; 2024-06-06 17:15:02 UTC; unix"
       }
     },
     "reprex": {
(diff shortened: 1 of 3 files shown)


jthompson-arcus:
It was also observed that "Include Suggests" would become enabled when it should be disabled. This was happening because the actually rendering would occur after the reactivity. A well documented issue when utilizing `renderUI()`. This is circumnavigated by using `freezeReactiveValue()`.

We should explore moving this outside of the shiny reactive environment. The current setup will cause the option to flicker to the enabled status briefly.

aclark02-arcus (approved):
LGTM! Waiting for write manifest workflow to successfully pass before merging.

manifest.json
```diff
@@ -4367,7 +4367,7 @@
         "Maintainer": "Kevin Ushey <kevin@rstudio.com>",
         "Repository": "RSPM",
         "Date/Publication": "2024-02-29 01:10:07 UTC",
-        "Built": "R 4.3.0; ; 2024-03-01 11:32:02 UTC; unix"
+        "Built": "R 4.3.3; ; 2024-06-06 17:15:02 UTC; unix"
```
jthompson-arcus, line 4370: Not sure why this changed

jthompson-arcus:
@aclark02-arcus The write manifest action was failing because it was installing a newer version of `rlang`. This was because we had the public rspm option selected so our workflow was ignoring the repo we set in the lock file.

aclark02-arcus (approved):
💪🏼

PR #792, opened 2024-06-17
Add user name to batch upload

### LINKED ISSUE #785: Attached username to batch upload decisions
Thinking about the batch upload process again. I feel like we need to include the username in the "Decision by" column on the database overview tab just in case an org wants to track that. So for example, it should read "Aaron (Batch Upload)". What do you think?

I was just thinking of a use case where there are several people able to run a "pkg approval process" outside of the app on new requested packages. Sometimes its done on large quantities of packages, but other times it may only be done on like 10 pkgs. It would be nice for us to know who uploaded those 10 pkgs to the app for auditing purposes.

Currently, the table looks like this, simply stating "Batch upload"

image

jthompson-arcus:
Closes #785

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: riskassessment
 Title: A web app designed to interface with the `riskmetric` package
-Version: 3.0.0.9024
+Version: 3.0.0.9025
 Authors@R: c( 
     person("Aaron", "Clark", role = c("aut", "cre"), email = "clark.aaronchris@gmail.com"),
     person("Robert", "Krajcik", role = "aut", email = "robert.krajcik@biogen.com"),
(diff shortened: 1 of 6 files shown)


jthompson-arcus:
User name is now added as a parenthetical to the batch upload designation.
image

aclark02-arcus (approved):
This looks good as-is. The only change I'd suggest is allowing users to make a decision when using `upload_pkg_list()` but I think that task can be saved for another day! I'll create a new issue for it. Thanks!

jthompson-arcus:
> This looks good as-is. The only change I'd suggest is allowing users to make a decision when using `upload_pkg_list()` but I think that task can be saved for another day! I'll create a new issue for it. Thanks!

Like if `is.null(user_name)` then just record it as `Batch Upload` instead of defaulting to `Batch Upload (system)`?

aclark02-arcus:
@Jeff-Thompson12, no. I mean't adding a new arg, called `decision` to that function. So it could be used like:

`upload_pkg_list(c("Tplyr","dplyr","dbplyr"), decision = c("Medium Risk", "Low Risk", ""))`

Just a suggestion.

PR #798, opened 2024-07-01
Update About

jthompson-arcus:


jthompson-arcus:
@aclark02-arcus I updated the team info data frame, the contact page, and the DESCRIPTION file. R packages can only have one maintainer, so I added co-lead as a comment and repositioned the entry.

Did I miss anything?

aclark02-arcus (approved):

PR #806, opened 2024-07-09
Correctly parse `riskmetric` dependency assessment

### LINKED ISSUE #805: Inaccurate display of number of dependencies
https://github.com/pharmaR/riskassessment/blob/a199a2eb8d0b7dae651c8d90df94e745e1d983f0/R/utils_insert_db.R#L205

If a package has 0 dependencies, `riskmetric` returns `character(0)` for the assessment value, but this code effectively renders it as `NA` which has a length of 1 instead of 0.

jthompson-arcus:
Closes #805

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,7 @@
 
 * Added navigation controls in Function Explorer tab (#644)
 * Fixed bug that crashed the Package Dependencies page for pkgs without any dependency info available (#802)
+* Fixed bug that incorrectly displayed 0 dependencies as 1 (#805)
 
 # riskassessment 3.1.0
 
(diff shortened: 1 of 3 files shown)


jthompson-arcus:
The change in metric calculation is more inline with how the [score is being calculated](https://pharmar.github.io/riskmetric/reference/metric_score.pkg_metric_dependencies.html) as well with using `NROW()`.

aclark02-arcus:
Was this PR supposed to synchronize `gower` too?

image

image

jthompson-arcus:
Yes, did you re-upload the package? It should be showing 0 dependencies.

aclark02-arcus:
> Yes, did you re-upload the package? It should be showing 0 dependencies.
> 
> You have to re-upload because the value is being written to the database.

Whoops, nope. For some reason, I was expecting the table to show one dep instead of the other way around (card showing 0 deps). Sorry

aclark02-arcus (approved):
💪🏼 
Looks good to me

PR #810, opened 2024-08-02
Remove evaluation recording from riskmetric assessments

jthompson-arcus:


jthompson-arcus:
@aclark02-arcus I really don't like this solution. I went down quite the rabbit hole to get here and further to look for a better answer, but I can't find one. We can just simply remove the `.recording` attribute and `with_eval_recording` class because the `vctrs` package is being used to create the `pkg_metric` objects. (Even if we could, that solution isn't very good either because it relies on these objects being relative immutable). The usage of the `vctrs` package effectively hijacks all inner assignment type calls and forces them to conform to a particular behavior which is why each assessment has to be rebuilt using `structure()`. If we want a none hacky solution to this problem, I think we are going to have to get it solved upstream in the `riskmetric` repo.

aclark02-arcus:
> we are going to have to get it solved upstream in the `riskmetric` repo

Ya, let's make Eric aware. For now, I will mark this as 'ready for review' and test it out later today.

aclark02-arcus (approved):

PR #828, opened 2025-01-09
Open card hyperlinks in a new tab

jthompson-arcus:


aclark02-arcus (approved):
Do you want this to hit our demo branch, presumably? Looks good. Merging.

aclark02-arcus (changes requested):
Would be nice to add this functionality to the html reports as well, if you don't mind adding it there too. Otherwise look good.

jthompson-arcus:
> Do you want this to hit our demo branch, presumably? Looks good. Merging.

Yeah, we can talk about it. I was thinking maybe a quick version `3.1.2` with subsequent demo branch.

aclark02-arcus (approved):

PR #834, opened 2025-01-23
Add `inherits` to list of used configs

jthompson-arcus:
The element `inherits` is a usable configuration option to inherit from another configuration in the same file. The app should not display a warning if it is used.


aclark02-arcus (approved):
Stellar work 💪🏼

PR #838, opened 2025-01-27
Fix bug where `Decision Automation` header isn't displaying for "less privileged" roles

aclark02-arcus:
I first noticed this bug in the webinar demo deployment:

https://rinpharma.shinyapps.io/riskassessment_shinygathering2025

I was logged in with the `reviewer` role, which means I didn't have the `auto_decision_adjust` privilege. As a result, I noticed that the "Decision Automation" header and horizontal line was missing from the `Upload Package` tab:

image

I checked the "normal" demo deployment, and this wasn't the case for the `reviewer` role, which was odd. The only difference in this deployment's automation rules is that it includes rules based on pkg scores. When I checked the code base, I saw that the `auto_decision_initial` depends on the existence of an upper/lower bound-based rule existing, which explained the issue. Thus, I just changed out the conditional to point to the logical value which monitors whether or not any risk rules exist.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,10 +2,11 @@
 
 Mostly minor changes here!
 
-* Open hyperlinks in a **new tab** from cards.
-* Fix bug causing app to crash with `ELSE` condition in rules
-* Add `inherits` to list of used configs
-* Incorporate bug fix for dependency assessment from `riskmetric` repo
+* Can now Open hyperlinks in a **new tab** from cards.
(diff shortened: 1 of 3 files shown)


aclark02-arcus:
Hey @jthompson-arcus, if you don't mind looking at this first thing in the morning, that would be great. Then I'll update my `webinardemo` deployment. Many thanks!

jthompson-arcus (approved):
Nice catch

